PR #12, opened 2015-04-20
changes to StripChromeSymbols.py (py3 support, __main__ guard)

ariccio:
I've added Python 3 support to StripChromeSymbols.py, and moved everything into main, guarded by an

```
if __name__ == "__main__":
```


ariccio:
NOTE: these are **unrelated** to the changes that I've discussed in the private email.

randomascii:
I merged your other StripChromeSymbols.py pull request, and this one seems to replicate some of the same changes. Can you rebase? It might be worth flattening the changes down to just one or two commits, to avoid creating a messy history.

ariccio:
> this one seems to replicate some of the same changes

D'oh! Silly me.

Will do.

> I merged your other StripChromeSymbols.py pull request

*Other* one? I think you've just self-evidentiated the case for a rebase.

PR #16, opened 2015-04-28
Test pull request

randomascii:
I want to see if creating a pull request will let me make comments on the individual lines of code.


UIforETW/About.h
```diff
@@ -16,20 +16,16 @@ limitations under the License.
 
 #pragma once
 
-// CAboutDlg dialog used for App About
-
-class CAboutDlg : public CDialogEx
+class CATLAboutDlg final : public ATL::CDialogImpl< CATLAboutDlg >
```
randomascii: Less code is good. Go ATL.

UIforETW/About.h
```diff
@@ -16,20 +16,16 @@ limitations under the License.
...
-};
+	BEGIN_MSG_MAP( CATLAboutDlg )
+		COMMAND_HANDLER(IDOK, BN_CLICKED, OnBnClickedOk)
+	END_MSG_MAP()
+	LRESULT OnBnClickedOk( WORD /*wNotifyCode*/, WORD /*wID*/, HWND /*hWndCtl*/, BOOL& /*bHandled*/ )
+	{
+		EndDialog( 1 );
+		return 1;
+	}
+};
```
randomascii: Note the missing end-of-line warning -- that should be fixed.

UIforETW/ChildProcess.cpp
```diff
@@ -29,7 +28,7 @@ ChildProcess::ChildProcess(std::wstring exePath)
 	// Create the pipe here so that it is guaranteed to be created before
 	// we try starting the process.
 	hPipe_ = CreateNamedPipe(kPipeName,
-		PIPE_ACCESS_DUPLEX | PIPE_TYPE_BYTE | PIPE_READMODE_BYTE,
+		(PIPE_ACCESS_DUPLEX | PIPE_TYPE_BYTE | PIPE_READMODE_BYTE),
```
randomascii: Why the parentheses? They're already on a separate line and separated by commas from other arguments. I guess I don't care either way, but it seems redundant.

UIforETW/ChildProcess.cpp
```diff
@@ -45,14 +44,14 @@ ChildProcess::~ChildProcess()
 {
 	if (hProcess_)
 	{
-		DWORD exitCode = GetExitCode();
+		const DWORD exitCode = GetExitCode();
```
randomascii: const is good.

UIforETW/ChildProcess.cpp
```diff
@@ -88,10 +87,13 @@ DWORD ChildProcess::ListenerThread()
 	if (ConnectNamedPipe(hPipe_, NULL) || GetLastError() == ERROR_PIPE_CONNECTED)
 	{
 		// Acquire the lock while writing to processOutput_
-		char buffer[1024];
+		const rsize_t bufferSize = 1024u;
+		char buffer[bufferSize];
 		DWORD dwRead;
 		while (ReadFile(hPipe_, buffer, sizeof(buffer) - 1, &dwRead, NULL) != FALSE)
 		{
+			if (dwRead >= bufferSize)
```
randomascii: That seems like an odd check. Harmless I guess, but it feels a bit like saying "if (false == true)" or "if (1 > 2)". The only way it could be possible is if Windows has a bug, or if the ReadFile line was changed. The former is implausible, the latter is a logic bug, and therefore assert() is appropriate.

UIforETW/ChildProcess.cpp
```diff
@@ -112,10 +114,11 @@ DWORD ChildProcess::ListenerThread()
 	return 0;
 }
 
-
+_Pre_satisfies_( hProcess_ == 0 )
+_Success_( return )
 bool ChildProcess::Run(bool showCommand, std::wstring args)
 {
-	assert(!hProcess_);
+	ATLASSERT(!hProcess_);
```
randomascii: What's the advantage of ATLASSERT? The default assert macro is pretty crappy (reentrancy problems in the UI) but is ATLASSERT any better?

UIforETW/ChildProcess.cpp
```diff
@@ -127,7 +130,7 @@ bool ChildProcess::Run(bool showCommand, std::wstring args)
 	if (hStdOutput_ == INVALID_HANDLE_VALUE)
 		return false;
 	if (!DuplicateHandle(GetCurrentProcess(), hStdOutput_, GetCurrentProcess(),
-		&hStdError_, 0, TRUE, DUPLICATE_SAME_ACCESS))
+		&hStdError_, 0, TRUE, DUPLICATE_SAME_ACCESS ))
```
randomascii: Delete extra space.

UIforETW/ChildProcess.cpp
```diff
@@ -140,18 +143,20 @@ bool ChildProcess::Run(bool showCommand, std::wstring args)
 	DWORD flags = CREATE_NO_WINDOW;
 	// Wacky CreateProcess rules say args has to be writable!
 	std::vector<wchar_t> argsCopy(args.size() + 1);
-	wcscpy_s(&argsCopy[0], argsCopy.size(), args.c_str());
-	BOOL success = CreateProcess(exePath_.c_str(), &argsCopy[0], NULL, NULL,
+	const int res = wcscpy_s(&argsCopy[0], argsCopy.size(), args.c_str());
```
randomascii: This is checking for an impossible error. Getting an error code would require both a bug in the buffer sizing logic *and* reconfiguring of wcscpy_s to return the error. wcscpy_s already has internal error checking so this adds no value.

It's sort of like going: "if (ptr) delete ptr;" - since 'delete' promises to do the 'if' check it is just bloat if user-code does it as well.

UIforETW/ChildProcess.cpp
```diff
@@ -214,14 +219,14 @@ void ChildProcess::WaitForCompletion(bool printOutput)
...
-		CloseHandle(hChildThread_);
+		handle_close::closeHandle(hChildThread_);
 		hChildThread_ = 0;
 	}
 
 	// Clean up.
 	if (hPipe_ != INVALID_HANDLE_VALUE)
 	{
-		CloseHandle(hPipe_);
+		handle_close::closeHandle(hPipe_);
```
randomascii: What is the advantage of handle_close::closeHandle() over just CloseHandle?

UIforETW/ChildProcess.cpp
```diff
@@ -17,8 +17,7 @@ limitations under the License.
 #include "stdafx.h"
 #include "ChildProcess.h"
 #include "Utility.h"
-
-#include <assert.h>
+#include "alias.h"
```
randomascii: Is this used? I didn't see it used anywhere.

UIforETW/Settings.h
```diff
@@ -54,8 +54,8 @@ class CSettings : public CDialogEx
 
 	CToolTipCtrl toolTip_;
 
-	virtual void DoDataExchange(CDataExchange* pDX);    // DDX/DDV support
-	virtual BOOL OnInitDialog();
+	virtual void DoDataExchange(CDataExchange* pDX) override;// DDX/DDV support
```
randomascii: override is good.

UIforETW/UIforETWDlg.cpp
```diff
@@ -433,27 +1196,47 @@ BOOL CUIforETWDlg::OnInitDialog()
 	return TRUE; // return TRUE unless you set the focus to a control
 }
 
-std::wstring CUIforETWDlg::GetDirectory(const wchar_t* env, const std::wstring& default)
+void CUIforETWDlg::initializeToolTip( )
```
randomascii: Having this in a separate function is good. But what purpose is served by having the text declared as variables? It seems to open up new potential for errors, it pollutes the (anonymous) global namespace, and it puts a huge separation between the definition of the strings and their usage. I'm not clear on the benefit.

UIforETW/UIforETWDlg.cpp
```diff
@@ -40,10 +41,821 @@ const int kRecordTraceHotKey = 1234;
...
+{
+	OutputDebugStringA("UIforETW: Encountered an unexpected error after calling GetFileAttributes!\r\n");
+	OutputDebugStringA("\tPath that we were checking: \r\n\t");
+	OutputDebugStringW(path);
+	OutputDebugStringA("\r\n");
+	ErrorHandling::outputErrorDebug(lastErr);
+	std::terminate();
+}
+
+//Compiler generates much better code when path is passed by reference.
```
randomascii: Passing strings by const reference is standard and doesn't need a comment.

UIforETW/UIforETWDlg.cpp
```diff
@@ -40,10 +41,821 @@ const int kRecordTraceHotKey = 1234;
...
+	OutputDebugStringA("UIforETW: Encountered an unexpected error after calling PathFileExists!\r\n");
+	OutputDebugStringA("\tPath that we were checking: \r\n\t");
+	OutputDebugStringW(path);
+	OutputDebugStringA("\r\n");
+	ErrorHandling::outputErrorDebug(lastErr);
+	std::terminate();
+}
+
+//Compiler didn't eliminate code for std::wstring
+//Seems to generate better code when passed a PCWSTR for path.
```
randomascii: Why are you micro-optimizing this? const ref string should be the standard. There is no need to optimize beyond what is clearest, safest, and most consistent.

UIforETW/UIforETWDlg.cpp
```diff
@@ -433,27 +1196,47 @@ BOOL CUIforETWDlg::OnInitDialog()
...
+	std::wstring result = getTraceDirFromEnvironmentVariable(env, default);
+	const bool doesFileExist = isValidPathToFSObject(result);
+	if (!doesFileExist)
 	{
-		(void)_wmkdir(result.c_str());
+		const bool sucessfullyCreatedDirectory =
+			enhancedCreateDirectory(result);
+		if (!sucessfullyCreatedDirectory)
+		{
+			AfxMessageBox(L"CUIforETWDlg::GetDirectory attempted to create a directory, but unexpectedly failed!");
```
randomascii: Should say what directory failed to be created in order to make the error message useful.

UIforETW/UIforETWDlg.cpp
```diff
@@ -40,10 +41,821 @@ const int kRecordTraceHotKey = 1234;
...
+{
+	//(first parameter):
+	//    There is a default string size limit for paths of 248 characters.
+	//    To extend this limit to 32,767 wide characters,
+	//        call the Unicode version of the function
+	//        and prepend "\\?\" to the path.
+	//                    ^ ==> L"\\\\?\\"
+
+	//TODO: what if network path?
+	//should we check if just starts with L"\\\\"?
```
randomascii: What is this doing? Why would the passed in path not be fine? What situation are you handling?

UIforETW/UIforETWDlg.cpp
```diff
@@ -40,10 +41,821 @@ const int kRecordTraceHotKey = 1234;
...
+	"SRV*c:\\symbols*http://msdl.microsoft.com/download/symbols";
+
+const char ChromiumSymbolPath[ ] =
+	"SRV*c:\\symbols*http://msdl.microsoft.com/download/symbols;"
+	"SRV*c:\\symbols*https://chromium-browser-symsrv.commondatastorage.googleapis.com";
+
+const char DefaultSymbolCachePath[ ] =
+	"c:\\symcache";
+
+void handleUnexpectedCreateDirectory( _In_ const std::wstring& path, _In_ const DWORD lastErr )
```
randomascii: These utility functions should go in utility.cpp to avoid cluttering the main source file - if they are needed at all.

UIforETW/UIforETWDlg.cpp
```diff
@@ -40,10 +41,821 @@ const int kRecordTraceHotKey = 1234;
...
+	std::terminate();
+}
+
+
+CRect GetWindowRectFromHwnd( _In_ const HWND hwnd )
+{
+	RECT windowRect_temp;
+	ASSERT(::IsWindow(hwnd));
+	const BOOL getWindowRectResult = ::GetWindowRect(hwnd, &windowRect_temp);
+	if (getWindowRectResult == 0)
```
randomascii: Don't handle impossible errors.

UIforETW/UIforETWDlg.cpp
```diff
@@ -40,10 +41,821 @@ const int kRecordTraceHotKey = 1234;
...
+{
+	if (!IsWindowsVistaOrGreater())
+	{
+		AfxMessageBox(L"ETW tracing requires Windows Vista or above.");
+		exit(10);
+	}
+	return;
+}
+
+void CheckDialogButtons(_In_ CWnd* const dlgToCheck, _In_ const bool bCompress,
```
randomascii: This has the same problem as the original tooltips button - it pretends to not be tightly coupled to the class, but it is. Having this as a non-member function means that it is more error prone and harder to update. It should be a member function.

UIforETW/UIforETWDlg.cpp
```diff
@@ -51,13 +863,16 @@ void outputPrintf(_Printf_format_string_ const wchar_t* pFormat, ...)
...
-void CUIforETWDlg::vprintf(const wchar_t* pFormat, va_list args)
+void CUIforETWDlg::vprintf(PCWSTR pFormat, va_list args)
 {
 	wchar_t buffer[5000];
 	_vsnwprintf_s(buffer, _TRUNCATE, pFormat, args);
 
-	for (const wchar_t* pBuf = buffer; *pBuf; ++pBuf)
+	for (PCWSTR pBuf = buffer; *pBuf; ++pBuf)
 	{
+		ATLASSERT(pBuf < (buffer + ARRAYSIZE(buffer)));
```
randomascii: You're checking for impossible things again. This condition can only happen if _vsnwprintf_s has failed, and if you start assuming that then you are programming on quicksand.

UIforETW/UIforETWDlg.cpp
```diff
@@ -191,14 +1010,14 @@ BEGIN_MESSAGE_MAP(CUIforETWDlg, CDialogEx)
 	ON_BN_CLICKED(IDC_SHOWCOMMANDS, &CUIforETWDlg::OnBnClickedShowcommands)
 	ON_BN_CLICKED(IDC_FASTSAMPLING, &CUIforETWDlg::OnBnClickedFastsampling)
 	ON_CBN_SELCHANGE(IDC_INPUTTRACING, &CUIforETWDlg::OnCbnSelchangeInputtracing)
-	ON_MESSAGE(WM_UPDATETRACELIST, UpdateTraceListHandler)
+	ON_MESSAGE(WM_UPDATETRACELIST, &CUIforETWDlg::UpdateTraceListHandler)
```
randomascii: Why?

UIforETW/UIforETWDlg.cpp
```diff
@@ -220,23 +1039,39 @@ END_MESSAGE_MAP()
...
 	{
-		bManageSymbolPath_ = true;
-		std::string symbolPath = "SRV*" + systemDrive_ + "symbols*http://msdl.microsoft.com/download/symbols";
-		if (bChromeDeveloper_)
-			symbolPath = "SRV*" + systemDrive_ + "symbols*http://msdl.microsoft.com/download/symbols;SRV*" + systemDrive_ + "symbols*https://chromium-browser-symsrv.commondatastorage.googleapis.com";
-		(void)_putenv(("_NT_SYMBOL_PATH=" + symbolPath).c_str());
-		outputPrintf(L"Setting _NT_SYMBOL_PATH to %s (Microsoft%s). "
-			L"Set _NT_SYMBOL_PATH yourself or toggle 'Chrome developer' if you want different defaults.\n",
-			AnsiToUnicode(symbolPath).c_str(), bChromeDeveloper_ ? L" plus Chrome" : L"");
+		OutputDebugStringA("UIforETW: _NT_SYMCACHE_PATH was found! "
```
randomascii: Should probably be OutputDebugStringW for consistency.

UIforETW/UIforETWDlg.cpp
```diff
@@ -244,96 +1079,58 @@ BOOL CUIforETWDlg::OnInitDialog()
...
+	hNotesAccelTable_  = loadAcceleratorsForEditingTraceNotes(instanceHandle);
+	hTracesAccelTable_ = loadAcceleratorsForActiveTraceList(instanceHandle);
+
+	CRect windowRect = GetWindowRectFromHwnd(m_hWnd);
 
-	CRect windowRect;
-	GetWindowRect(&windowRect);
-	initialWidth_ = lastWidth_ = windowRect.Width();
-	initialHeight_ = lastHeight_ = windowRect.Height();
+	initialWidth_ = windowRect.Width();
```
randomascii: I'm not saying that this alternate form is worse, but is it clearly better? Churn has to be justified.

UIforETW/UIforETWDlg.cpp
```diff
@@ -40,10 +41,821 @@ const int kRecordTraceHotKey = 1234;
...
+	if (getWindowRectResult == 0)
+	{
+		outputPrintf(L"GetWindowRect failed!!\n");
+		ErrorHandling::outputErrorDebug();
+		std::terminate();
+	}
+	return windowRect_temp;
+}
+
+void SetSaveTraceBuffersWindowText( _In_ const HWND hWnd )
```
randomascii: Don't handle impossible errors. What *value* does this add?

UIforETW/UIforETWDlg.cpp
```diff
@@ -624,36 +1411,56 @@ std::wstring CUIforETWDlg::GenerateResultFilename() const
...
+	const bool validTimeStr = (3 == sscanf_s(time, "%d:%d:%d", &hour, &min, &sec));
+#pragma warning(suppress: 6054)
+	const bool validDateStr = (3 == sscanf_s(date, "%d/%d/%d", &month, &day, &year));
+
+	std::wstring filePart;
+	if (validTimeStr && validDateStr)
 	{
 		// The filenames are chosen to sort by date, with username as the LSB.
-		swprintf_s(fileName, L"%04d-%02d-%02d_%02d-%02d-%02d_%s", year + 2000, month, day, hour, min, sec, username);
+		const int res = swprintf_s(fileName, L"%04d-%02d-%02d_%02d-%02d-%02d_%s", year + 2000, month, day, hour, min, sec, username);
```
randomascii: Don't handle impossible errors. swprintf_s *will not* return an error code. It will fail if the text won't fit.

UIforETW/UIforETWDlg.cpp
```diff
@@ -40,10 +41,821 @@ const int kRecordTraceHotKey = 1234;
...
+		}
+		return false;
+	}
+	if ((pathAttributes & FILE_ATTRIBUTE_DIRECTORY) == 0)
+		return false;
+	//all tests passed!
+	return true;
+}
+
+std::wstring getKernelStackWalk( _In_ const bool bSampledStacks, _In_ const bool bCswitchStacks )
```
randomascii: I understand the desire to not have a monster function that sets all the profile settings, but is this really the best way to break it up? It seems like it actually reduces clarity by spreading the code out too much.

UIforETW/UIforETWDlg.cpp
```diff
@@ -40,10 +41,821 @@ const int kRecordTraceHotKey = 1234;
...
+	if (setSymbolPathResult == 0)
+	{
+		outputPrintf(L"Failed to set an environment variable!\n\t"
+					 L"Attempted to set environment variable: `%S`\n\t"
+					 L"Value that we attempted to set the variable to: `%S`\n",
+					 variableName, value);
+		ErrorHandling::outputPrintfErrorDebug();
+		return false;
+	}
+	OutputDebugStringA("UIforETW: Successfully set an environment variable.\r\n\t`");
```
randomascii: What is the point of having all the OutputDebugStringA calls?

UIforETW/UIforETWDlg.cpp
```diff
@@ -1101,7 +1909,8 @@ void CUIforETWDlg::OnLbnDblclkTracelist()
 {
 	int selIndex = btTraces_.GetCurSel();
 	// This check shouldn't be necessary, but who knows?
-	if (selIndex < 0 || selIndex >= (int)traces_.size())
+	//That comment IS WRONG! CListBox::GetCurSel MAY return -1! (LB_ERR, when no item is selected)
```
randomascii: It is true that in general GetCurSel can return -1, but it cannot do this in OnLbnDblclkTracelist, which is called when the user double-clicks an item in the list, an it?

UIforETW/UIforETWDlg.cpp
```diff
@@ -1131,7 +1940,7 @@ void CUIforETWDlg::OnSize(UINT nType, int /*cx*/, int /*cy*/)
 		int yDelta = windowRect.Height() - lastHeight_;
 		lastHeight_ += yDelta;
 
-		UINT flags = SWP_NOMOVE | SWP_NOOWNERZORDER | SWP_NOZORDER | SWP_NOACTIVATE;
+		const UINT flags = (SWP_NOMOVE | SWP_NOOWNERZORDER | SWP_NOZORDER | SWP_NOACTIVATE);
```
randomascii: const is good. I don't understand the value behind the parentheses.

UIforETW/UIforETWDlg.cpp
```diff
@@ -1157,9 +1966,7 @@ void CUIforETWDlg::SaveNotesIfNeeded()
 	if (editedNotes != traceNotes_)
 	{
 		if (!traceNoteFilename_.empty())
-		{
 			WriteTextAsFile(traceNoteFilename_, editedNotes);
```
randomascii: Deleting the parentheses is a style thing of *extremely* dubious value. Churn.

UIforETW/UIforETWDlg.cpp
ariccio: It's possible, just don't click *an item* in the list.

https://www.youtube.com/watch?v=MiMaXdGbfOc

UIforETW/ChildProcess.cpp
ariccio: In retrospect, this is totally nonsensical. An assert might make sense - not sure what I was thinking.

UIforETW/ChildProcess.cpp
ariccio: - `assert( )` calls abort (via _wassert), and breaks in the middle of the CRT, which means that you have to find the spot where the assertion actually fired (small deal, but annoying). It also means that hitting `Ignore` doesn't actually let the program continue.
- `ATLASSERT( )`, which is defined as `#define ATLASSERT(expr) _ASSERTE(expr)`, calls `_CrtDbgReport( )` and then `_CrtDbgBreak( )`, which allows the program to continue, and also breaks right into your code.
- `MFC`'s `ASSERT( )` is in between the two.


Kinda related, but interesting reading about ASSERT:
- [Press Cancel to cancel (or Cancelable Asserts)](https://lanzkron.wordpress.com/2013/10/02/press-cancel-to-cancel-or-cancelable-asserts/)

UIforETW/ChildProcess.cpp
ariccio: > wcscpy_s already has internal error checking so this adds no value.

The VS2013 headers are messy enough (`#define _FUNC_NAME wcscpy_s` - *really, are you kidding me???!?*) that I'll take your word for it.

> It's sort of like going: "if (ptr) delete ptr;" - since 'delete' promises to do the 'if' check it is just bloat if user-code does it as well.

Ahh, I hit your pet-peeve :)

UIforETW/UIforETWDlg.cpp
randomascii: Huh. Fascinating. I had no idea that was possible. Well, good thing I had the checking, and that comment definitely needs removing or changing.

UIforETW/UIforETWDlg.cpp
ariccio: > Why are you micro-optimizing this?

Compulsion, really. No *real* logical reason.

I remember looking at the generated assembly, and noticing that the compiler had generated all sorts of code for throwing exceptions and constructing strings, and it all seemed very silly for something that I was going to just pass to OutputDebugStringW.

>  There is no need to optimize beyond what is clearest, safest, and most consistent.

Clarity always (should) trump micro-optimizations, so this is a mistake on my part.

UIforETW/UIforETWDlg.cpp
ariccio: Fair enough, although I don't think this is an *impossible* error.

TLDR: Today I'm running with -Wpedantic, -Wnitpick, and -Wannoying. Will remove this code.

* It can happen, and I found [one (if only one) instance of problems caused by not checking the return value](http://forums.codeguru.com/printthread.php?t=305640)
* There is a minute chance of a race-condition, where the window is destroyed (and thus the handle is invalid) before `GetWindowRect` is called - in which case maybe we should mark the function with something like `_No_competing_thread_`? (strangely, that annotation isn't documented)
* Unlike `wcscpy_s`, we don't have the source code, so by assuming that failure is impossible is technically an implementation detail, and as a longtime reader of Raymond Chen's blog, I have ethical issues with relying on implementation details.

ariccio: This'd be a good place to use `VERIFY( )`

UIforETW/UIforETWDlg.cpp
ariccio: Mostly the same reasons as for GetWindowRect (above)



...but I looked at the docs again, and noticed this in the comments:
setwindowtextcomments

I'll have to investigate that.

UIforETW/UIforETWDlg.cpp
ariccio: Ok.

- [ ] Fixed

UIforETW/UIforETWDlg.cpp
randomascii: Since the one likely way it could fail would be due to being called after the window is destroyed, which would be a bug in UIforETW, it makes sense to use Verify(). If it fails for other reasons then I don't know how to handle it - I have to assume that any error handling code will fail also - so there is no *value* in trying to handle it.

UIforETW/UIforETWDlg.cpp
randomascii: It is reasonable for WIndows to require that strings be natively aligned (aligned to sizeof(TCHAR)), so again that would be a UIforETW logic error if SetWindowText started failing. The comment saying that 8 or 16-byte alignment is required is almost certainly wrong. No, I'm going to go with certainly wrong. I think they meant 8 or 16 *bit* alignment.

ariccio: I'm sold. Will do.

UIforETW/UIforETWDlg.cpp
ariccio: Best way? Probably not. This was the result of me trying to break up the code that I can experiment with it.

It's on it's way to `/dev/null`.

UIforETW/UIforETWDlg.cpp
ariccio: You know, I think "programming on quicksand" actually describes everything that's wrong with my current style of programming. It's like, hilariously accurate.

randomascii: :-) I think it was Steve McConnell who wrote about the advantages of no-fail programming. He gave the example of asking various candidates to write a toupper function. Many of them incorporated error handling. But as soon as you do that - as soon as core functions like that can fail - then the necessary error handling becomes an exponential explosion of (untested) code paths. How much better it is to write a toupper that always works.

Of course, some functions can fail, and that may have to be handled. But having as few functions as possible that can fail, and knowing (or deciding, or mandating) what you can count on is very powerful. Learning how to compose large systems with as little error handling code as possible, while still handling all possible errors, is an art form.

Video games offer one example. They hook allocation functions at a low-level so that they will never return NULL or throw an exception. If an allocation fails then it is a fatal error and it is handled (deep inside the alloc function) with __debugbreak() or exit() or some-such. The error is handled, but at zero size and complexity cost in the client code.

And thus quicksand is avoided. This technique does not apply to all software, or at least not to all allocations in all software (if a user can specify an arbitrarily large image size then the application should handle a failure to allocate the image buffer) but it is powerful.

ariccio: It's  funny that you mention allocation functions/large image buffers, because I remember a problem in chrome that I hit a while ago, where code depended on `base::UncheckedMalloc` returning a `NULL` if it failed to allocate.

On Windows, when `malloc` fails (`base::UncheckedMalloc` just calls `malloc`), the whole process terminates, which is ***EXACTLY THE OPPOSITE*** of what the documentation for `base::UncheckedMalloc` says.

Thus, `SkMallocPixelRef` would try to allocate an absurdly large buffer, and the whole browser would explode.

https://code.google.com/p/chromium/issues/detail?id=360681

/digression

Steve McConnell's books are now on my mental "to read" list. Sounds interesting.

UIforETW/UIforETWDlg.cpp
ariccio: - bitwise operators have strange precedence rules
- primary expressions are more obviously correct/easier for the reader to reason about
- parenthesis are cheap

...other than that, they're unnecessary.

UIforETW/ChildProcess.cpp
ariccio: Same as https://github.com/google/UIforETW/pull/16/files#r29552320

UIforETW/UIforETWDlg.cpp
randomascii: The parentheses don't seem to add any clarity (to my eye) in this case. My recommendation for borderline style changes would be to not make them, because their value is probably less than the cost of the churn. On the other hand, if you wrote a new line of code with the parentheses I wouldn't change it, for the same churn-avoidance reasons.

That is, if a change has a clear benefit (whatever *that* means) then make it. If it is style issue where reasonable people would disagree then avoiding churn wins.

I don't know if that explanation is any use since it depends completely on a subject understanding of clarity, value, style, etc., but I thought I'd share my thinking.

ariccio: I think the relevant definition of *clear* here is not whether something has a benefit that's obvious to the person proposing the change, but whether that benefit is *unanimously seen*.

Thus, by your observation, this change is *ipso facto* not worth the churn. Sharing thoughts is always a good idea.

UIforETW/UIforETWDlg.cpp
ariccio: > parentheses

Braces?

The funny thing here, is that I'm usually very much so in favor of using braces in **every single conditional**, but here I decided to follow [Google style](https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Conditionals) fair enough.

UIforETW/ChildProcess.cpp
ariccio: static analysis picks up misuse - code like:

```
handle_close::closeHandle(someHandle);
```
(several hundred lines of code in between)
```
handle_close::closeHandle(someHandle);
```
triggers a code analysis warning/error

randomascii: That sucks that the regular CloseHandle doesn't trigger the same warnings. I'll have to experiment with that.

ariccio: Yeah, it's the `_Post_ptr_invalid_` hack that I mentioned in an email a while back

ariccio: For your convenience, here's the email, with the part about the annotation:

> [....]I came up with the idea of wrapping CloseHandle with a function like this:
```
_Success_( return )
BOOL safeCloseHandle( _Pre_valid_ _Post_ptr_invalid_ HANDLE handle_to_close )
{
	return CloseHandle( handle_to_close );
}
```
It was after I made the connection that a HANDLE is just a `typedef`d `void*`, and thus it is "invalidated" in a sense by calling CloseHandle. It works beautifully in contrived examples; if you try to call safeCloseHandle (or a better name`*`) twice, then it is reliably picked up by /analyze, with a nice "using uninitialized memory" error. I've been planning to try it on a large codebase since then, particularly Notepad++, but Notepad++ has a fairly slow development cycle. I've also been planning an article on it, so trying it on a large project would be a neat validation of the idea.

> So I decided a few minutes ago to try it on UIforETW, just to see how it'd work. There are only ~10 calls to CloseHandle, so I didn't expect any bugs (I didn't find any), but safeCloseHandle did work exactly as it did in smaller examples.

> If you open up [the toy-around commit that I pushed](https://github.com/ariccio/main/commit/8d9a91cfa4f211be61e225e7b435408308920f7a), uncomment any of the duplicated calls to safeCloseHandle, and rebuild, /analyze catches it. Comment out the duplicate safeCloseHandle calls, add some duplicate CloseHandle calls, and rebuild, no complaint from /analyze. I think it's pretty cool.

> `*safeCloseHandle` is probably a bad name, as it gives a "positive" illusion of safety, and thus people will abuse/misinterpret it. It should really be something like "checkedCloseHandle", or "checked_wrappers:close_handle".

randomascii: What about the other alternative of having a HANDLE wrapper class? This
could easily make it impossible to make redundant calls to CloseHandle(),
and impossible bugs are even better than bugs that the compiler warns about.

On Sun, May 3, 2015 at 11:00 PM, Alexander Riccio <notifications@github.com>
wrote:

> In UIforETW/ChildProcess.cpp
> <https://github.com/google/UIforETW/pull/16#discussion_r29567178>:
>
> >  		hChildThread_ = 0;
> >  	}
> >
> >  	// Clean up.
> >  	if (hPipe_ != INVALID_HANDLE_VALUE)
> >  	{
> > -		CloseHandle(hPipe_);
> > +		handle_close::closeHandle(hPipe_);
>
> For your convenience, here's the email, with the part about the annotation:
>
> [....]I came up with the idea of wrapping CloseHandle with a function like
> this:
>
> _Success_( return )
> BOOL safeCloseHandle( _Pre_valid_ _Post_ptr_invalid_ HANDLE handle_to_close )
> {
>     return CloseHandle( handle_to_close );
> }
>
> It was after I made the connection that a HANDLE is just a typedefd void*,
> and thus it is "invalidated" in a sense by calling CloseHandle. It works
> beautifully in contrived examples; if you try to call safeCloseHandle (or a
> better name*) twice, then it is reliably picked up by /analyze, with a
> nice "using uninitialized memory" error. I've been planning to try it on a
> large codebase since then, particularly Notepad++, but Notepad++ has a
> fairly slow development cycle. I've also been planning an article on it, so
> trying it on a large project would be a neat validation of the idea.
>
> So I decided a few minutes ago to try it on UIforETW, just to see how it'd
> work. There are only ~10 calls to CloseHandle, so I didn't expect any bugs
> (I didn't find any), but safeCloseHandle did work exactly as it did in
> smaller examples.
>
> If you open up the toy-around commit that I pushed
> <https://github.com/ariccio/main/commit/8d9a91cfa4f211be61e225e7b435408308920f7a>,
> uncomment any of the duplicated calls to safeCloseHandle, and rebuild,
> /analyze catches it. Comment out the duplicate safeCloseHandle calls, add
> some duplicate CloseHandle calls, and rebuild, no complaint from /analyze.
> I think it's pretty cool.
>
> *safeCloseHandle is probably a bad name, as it gives a "positive"
> illusion of safety, and thus people will abuse/misinterpret it. It should
> really be something like "checkedCloseHandle", or
> "checked_wrappers:close_handle".
>
> —
> <https://github.com/google/UIforETW/pull/16/files#r29567178>.
>



-- 
Bruce Dawson

randomascii:
Closing, since this was just a test pull request.

ariccio:
*feigns surprise*

PR #17, opened 2015-04-28
Swapped MFC about dialog for ATL, removed unused includes

ariccio:
The ATL-style about box code is much simpler, and ATL-style code tends
to be smaller & more efficient. I added `#include <atlwin.h>` for this,
and `#define`d `_ATL_NO_AUTOMATIC_NAMESPACE` to prevent weird namespace
pollution issues. I removed the close button on the about box - because
*what's the point of it?*

I can't build unless `<afxwin.h>` is the VERY FIRST `#include` - some
other people have this problem with MFC at random - so I moved that
first.

I've also removed some unused includes (MFC support for *IE4 common
controls*, MFC automation).


ariccio:
That must be because I deleted `About.cpp`?
huh

randomascii:
Okay, merged. The conflict was in stdafx.h. It was a conflict between your changes to that file.

It is probably worth redoing these changes (copying them to a clean repo or manually merging them) because I ended up with the last twelve changes duplicated in the history, which is just messy. It may be possible to avoid this within git, but it would be simpler (for me anyway) if future pull requests were just one change away from tip-of-tree.

ariccio:
Oops, yeah I need to fix the remote tracking config in git -- my local copy of `master` wasn't showing up as behind tip-of-the-tree >:(

PR #20, opened 2015-05-03
added override in 5 places

ariccio:
`override` is always good.

### DIFF
--- a/UIforETW/Settings.h
+++ b/UIforETW/Settings.h
@@ -54,8 +54,8 @@ protected:
 
 	CToolTipCtrl toolTip_;
 
-	virtual void DoDataExchange(CDataExchange* pDX);    // DDX/DDV support
-	virtual BOOL OnInitDialog();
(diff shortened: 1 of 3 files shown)


randomascii:
I'd love to have a warning that triggered when 'override' was missing. I'd turn that on in a heartbeat, although I guess the MFC headers would trigger it a million times. Anyway, good change.

ariccio:
I believe that [C4263](https://msdn.microsoft.com/en-us/library/ay4h0tc9.aspx), [C4264](https://msdn.microsoft.com/en-us/library/tb55b5f4.aspx), [C4265](https://msdn.microsoft.com/en-us/library/wzxffy8c.aspx), and [C4266](https://msdn.microsoft.com/en-us/library/4b76ty10.aspx), do a half-assed job at warning where override should be used.


...and if I remember correctly, the MFC headers DO trigger it a few million times. Working around it requires a ton of ugly `#pragma warning(disable, ....)`s, and `#pragma warning(push)`/`#pragma warning(pop)` combos.

The MFC headers do a lot of really questionable things (as I've mentioned before).

PR #22, opened 2015-05-03
Python cleanup & minor refactoring

ariccio:
I improved [style compliance](https://www.chromium.org/chromium-os/python-style-guidelines) of the Python scripts, and also improved the Python 2.x/3.x cross-compatibility. Running pylint helps!

Removed a few usages of depreciated functions, `os.popen` & `has_key`, and replaced with modern/pythonic alternatives.

For now, I couldn't remove a few globals from XperfProcessParentage.py. Grr.

This should improve readability.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@
 *Generated*
 # Excluded most executables, but include a few specific ones.
 *.exe
-
+*.pyc
 *.ilk
 *.pdb
 Debug/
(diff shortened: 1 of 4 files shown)


randomascii:
Looks good, overall. Thanks for the cleanup work. It's too bad that the github diffing algorithms don't do a better job of comparing code when it gets indented (as when it is moved into main()), but oh well.

There's one tricky change that I need which is removing the expletives regarding global variables. This is only tricky because I don't want them to be in the history either, so some commit merging or revision or manual copying will be needed. Thanks.

ariccio:
[This isn't the Linux kernel](http://www.vidarholen.net/contents/wordcount/).
Fair enough.

PR #25, opened 2015-05-07
changed `assert`s to `ATLASSERT`s

ariccio:
I also inserted `_Pre_satisfies_` annotations for function and object/member function preconditions.

In `Settings.cpp`, I changed an implicit conversion of `*CSettings`/`*this` to `HWND` (via `CWnd::operator HWND`) to explicitly passing m_hWnd.

`ATLASSERT` calls `_CrtDbgBreak` instead of `abort`, and is generally suited better to UI programming.

### DIFF
--- a/UIforETW/ChildProcess.cpp
+++ b/UIforETW/ChildProcess.cpp
@@ -114,7 +114,7 @@ DWORD ChildProcess::ListenerThread()
 _Pre_satisfies_(!(this->hProcess_))
 bool ChildProcess::Run(bool showCommand, std::wstring args)
 {
-	ATLASSERT(!hProcess_);
+	UIETWASSERT(!hProcess_);
 
 	if (showCommand)
 		outputPrintf(L"%s\n", args.c_str());
--- a/UIforETW/ChildProcess.cpp
+++ b/UIforETW/ChildProcess.cpp
@@ -112,10 +112,10 @@ DWORD ChildProcess::ListenerThread()
 	return 0;
 }
 
-
+_Pre_satisfies_(!(this->hProcess_))
 bool ChildProcess::Run(bool showCommand, std::wstring args)
 {
-	assert(!hProcess_);
+	ATLASSERT(!hProcess_);
 
 	if (showCommand)
 		outputPrintf(L"%s\n", args.c_str());
--- a/UIforETW/ChildProcess.cpp
+++ b/UIforETW/ChildProcess.cpp
@@ -18,7 +18,6 @@ limitations under the License.
 #include "ChildProcess.h"
 #include "Utility.h"
 
-#include <assert.h>
 #include <vector>
 
 static const wchar_t* kPipeName = L"\\\\.\\PIPE\\UIforETWPipe";
--- a/UIforETW/DirectoryMonitor.cpp
+++ b/UIforETW/DirectoryMonitor.cpp
@@ -16,7 +16,6 @@ limitations under the License.
 
 #include "stdafx.h"
 #include "DirectoryMonitor.h"
-#include <assert.h>
 
 DirectoryMonitor::DirectoryMonitor(CWnd* pMainWindow)
 	: mainWindow_(pMainWindow)
--- a/UIforETW/DirectoryMonitor.cpp
+++ b/UIforETW/DirectoryMonitor.cpp
@@ -37,7 +37,7 @@ DWORD DirectoryMonitor::DirectoryMonitorThread()
 
 	if (hChangeHandle == INVALID_HANDLE_VALUE)
 	{
-		ATLASSERT(0);
+		UIETWASSERT(0);
 		return 0;
 	}
 
@@ -52,7 +52,7 @@ DWORD DirectoryMonitor::DirectoryMonitorThread()
 			mainWindow_->PostMessage(WM_UPDATETRACELIST, 0, 0);
 			if (FindNextChangeNotification(hChangeHandle) == FALSE)
 			{
-				ATLASSERT(0);
+				UIETWASSERT(0);
 				return 0;
 			}
 			break;
@@ -61,7 +61,7 @@ DWORD DirectoryMonitor::DirectoryMonitorThread()
 			return 0;
 
 		default:
-			ATLASSERT(0);
+			UIETWASSERT(0);
 			break;
 		}
 	}
@@ -72,8 +72,8 @@ _Pre_satisfies_(this->hThread_ == 0)
 _Pre_satisfies_(this->hShutdownRequest_ == 0)
 void DirectoryMonitor::StartThread(const std::wstring* traceDir)
 {
-	ATLASSERT(hThread_ == 0);
-	ATLASSERT(hShutdownRequest_ == 0);
+	UIETWASSERT(hThread_ == 0);
+	UIETWASSERT(hShutdownRequest_ == 0);
 	traceDir_ = traceDir;
 	// No error checking -- what could go wrong?
 	hThread_ = CreateThread(nullptr, 0, DirectoryMonitorThreadStatic, this, 0, 0);
--- a/UIforETW/DirectoryMonitor.cpp
+++ b/UIforETW/DirectoryMonitor.cpp
@@ -38,7 +38,7 @@ DWORD DirectoryMonitor::DirectoryMonitorThread()
 
 	if (hChangeHandle == INVALID_HANDLE_VALUE)
 	{
-		assert(0);
+		ATLASSERT(0);
 		return 0;
 	}
 
@@ -53,7 +53,7 @@ DWORD DirectoryMonitor::DirectoryMonitorThread()
 			mainWindow_->PostMessage(WM_UPDATETRACELIST, 0, 0);
 			if (FindNextChangeNotification(hChangeHandle) == FALSE)
 			{
-				assert(0);
+				ATLASSERT(0);
 				return 0;
 			}
 			break;
@@ -62,21 +62,23 @@ DWORD DirectoryMonitor::DirectoryMonitorThread()
 			return 0;
 
 		default:
-			assert(0);
+			ATLASSERT(0);
 			break;
 		}
 	}
   // Unreachable.
 }
 
+_Pre_satisfies_(this->hThread_ == 0)
+_Pre_satisfies_(this->hShutdownRequest_ == 0)
 void DirectoryMonitor::StartThread(const std::wstring* traceDir)
 {
-	assert(hThread_ == 0);
-	assert(hShutdownRequest_ == 0);
+	ATLASSERT(hThread_ == 0);
+	ATLASSERT(hShutdownRequest_ == 0);
 	traceDir_ = traceDir;
 	// No error checking -- what could go wrong?
 	hThread_ = CreateThread(nullptr, 0, DirectoryMonitorThreadStatic, this, 0, 0);
-	hShutdownRequest_ = CreateEvent(nullptr, FALSE, FALSE, nullptr);;
+	hShutdownRequest_ = CreateEvent(nullptr, FALSE, FALSE, nullptr);
 }
 
 DirectoryMonitor::~DirectoryMonitor()
--- a/UIforETW/Settings.cpp
+++ b/UIforETW/Settings.cpp
@@ -145,7 +145,7 @@ void CSettings::OnBnClickedCopystartupprofile()
 	wchar_t documents[MAX_PATH];
 
 	const BOOL getMyDocsResult = SHGetSpecialFolderPathW(m_hWnd, documents, CSIDL_MYDOCUMENTS, TRUE);
-	ATLASSERT(getMyDocsResult);
+	UIETWASSERT(getMyDocsResult);
 	if (!getMyDocsResult)
 	{
 		OutputDebugStringA("Failed to find My Documents directory.\r\n");
--- a/UIforETW/Settings.cpp
+++ b/UIforETW/Settings.cpp
@@ -143,11 +143,15 @@ void CSettings::OnBnClickedCopystartupprofile()
 	std::wstring source = exeDir_ + fileName;
 
 	wchar_t documents[MAX_PATH];
-	if (!SHGetSpecialFolderPath(*this, documents, CSIDL_MYDOCUMENTS, TRUE))
+
+	const BOOL getMyDocsResult = SHGetSpecialFolderPathW(m_hWnd, documents, CSIDL_MYDOCUMENTS, TRUE);
+	ATLASSERT(getMyDocsResult);
+	if (!getMyDocsResult)
 	{
-		assert(!"Failed to find My Documents directory.\n");
+		OutputDebugStringA("Failed to find My Documents directory.\r\n");
 		return;
 	}
+
 	std::wstring dest = documents + std::wstring(L"\\WPA Files\\") + fileName;
 	if (CopyFile(source.c_str(), dest.c_str(), FALSE))
 	{
--- a/UIforETW/UIforETWDlg.cpp
+++ b/UIforETW/UIforETWDlg.cpp
@@ -271,16 +271,16 @@ BOOL CUIforETWDlg::OnInitDialog()
 	// Add "About..." menu item to system menu.
 
 	// IDM_ABOUTBOX must be in the system command range.
-	ASSERT((IDM_ABOUTBOX & 0xFFF0) == IDM_ABOUTBOX);
-	ASSERT(IDM_ABOUTBOX < 0xF000);
+	ATLASSERT((IDM_ABOUTBOX & 0xFFF0) == IDM_ABOUTBOX);
+	ATLASSERT(IDM_ABOUTBOX < 0xF000);
 
 	CMenu* pSysMenu = GetSystemMenu(FALSE);
 	if (pSysMenu)
 	{
-		BOOL bNameValid;
+		//BOOL bNameValid;
 		CString strAboutMenu;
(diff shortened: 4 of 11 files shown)


UIforETW/Settings.cpp
```diff
@@ -143,11 +143,15 @@ void CSettings::OnBnClickedCopystartupprofile()
 	std::wstring source = exeDir_ + fileName;
 
 	wchar_t documents[MAX_PATH];
-	if (!SHGetSpecialFolderPath(*this, documents, CSIDL_MYDOCUMENTS, TRUE))
+
+	const BOOL getMyDocsResult = SHGetSpecialFolderPathW(m_hWnd, documents, CSIDL_MYDOCUMENTS, TRUE);
```
randomascii: I'm not convinced that accessing a protected member variable of the base class directly is better than a conversion operator. But mostly I'm annoyed that m_hWnd is a protected member variable. So much for encapsulation.

At least using the conversion operator doesn't advertise the fact that m_hWnd is completely accessible, and the HWND conversion operator is sure safer than, say, the CString conversion operator to const char*.

UIforETW/UIforETWDlg.cpp
```diff
@@ -271,16 +271,16 @@ BOOL CUIforETWDlg::OnInitDialog()
...
-	ASSERT((IDM_ABOUTBOX & 0xFFF0) == IDM_ABOUTBOX);
-	ASSERT(IDM_ABOUTBOX < 0xF000);
+	ATLASSERT((IDM_ABOUTBOX & 0xFFF0) == IDM_ABOUTBOX);
+	ATLASSERT(IDM_ABOUTBOX < 0xF000);
 
 	CMenu* pSysMenu = GetSystemMenu(FALSE);
 	if (pSysMenu)
 	{
-		BOOL bNameValid;
+		//BOOL bNameValid;
```
randomascii: Don't leave commented out code unless it is needed for documentation purposes.

UIforETW/ChildProcess.cpp
```diff
@@ -112,10 +111,10 @@ DWORD ChildProcess::ListenerThread()
 	return 0;
 }
 
-
+_Pre_satisfies_(!(this->hProcess_))
 bool ChildProcess::Run(bool showCommand, std::wstring args)
 {
-	assert(!hProcess_);
+	ATLASSERT(!hProcess_);
```
randomascii: Lets take it as a given that ATLASSERT is enough better than assert to justify the churn of changing all instances. Even with assumption, is ATLASSERT perfect? Probably not. So why not change ATLASSERT to UIETWASSERT or some-such, and #define that as ATLASSERT. Then if we ever want to change it back, or to an assert that is actually good, it won't require any churn.

Just a thought. Not guaranteed to be a good on.

UIforETW/DirectoryMonitor.cpp
```diff
@@ -62,21 +61,23 @@ DWORD DirectoryMonitor::DirectoryMonitorThread()
...
 		default:
-			assert(0);
+			ATLASSERT(0);
 			break;
 		}
 	}
   // Unreachable.
 }
 
+_Pre_satisfies_(this->hThread_ == 0)
```
randomascii: It's too bad these annotations are ugly. I definitely don't want hundreds of annotations because that gets unmaintainable and then you end up with buggy annotations and unreadable code.

UIforETW/UIforETWDlg.cpp
randomascii: But definitely a clear improvement.

randomascii:
Seems fairly reasonable, but I made a few comments.

UIforETW/DirectoryMonitor.cpp
ariccio: > It's too bad these annotations are ugly

[Don't hate them because they're beautiful](https://www.osr.com/blog/2015/02/23/sal-annotations-dont-hate-im-beautiful/).

UIforETW/Settings.cpp
ariccio: > But mostly I'm annoyed that m_hWnd is a protected member variable. So much for encapsulation.

On the other hand, what would "encapsulation" look like here? `HWND getHWND() const;`? What would that encapsulate? 

Personally, I think that simple getters & setters aren't really encapsulation, and where they're really needed, it's a sign that objects are too highly coupled, and far too big.

Maybe it's a bit unrealistic - I really don't know. Maybe it's just my inner C programmer speaking.

> At least using the conversion operator doesn't advertise the fact that m_hWnd is completely accessible

Well, yeah it doesn't advertise that fact, but it sure is surprising that a `CSettings` object is implicitly convertible to something else, partly because it doesn't advertise anything.

> CString conversion operator to const char*.

`CString` is pretty horrible, in all sorts of ways. No move constructors, passing by value does expensive reference counting, weird buffer locking/unlocking mechanisms, generally bad memory management, etc..... I've really developed a deep dislike of it.

----what follows is a bit of a rant, sorry!----
The best example of a good use of encapsulation that comes to mind is something from my work on WinDirStat.

"Child" objects were heap allocated and stored in a vector of pointers (i.e. `std::vector<CItem*>`), and those items were displayed in a list view, sorted by size, date, or name (etc...).

When I changed everything to allocate an array of objects and use placement new to construct them (which saved a lot of space, and was faster),  I couldn't sort them anymore, because then I'd actually have to move the location of the objects in memory, which would require fixing a ton of pointers to point to the new location.

Instead of crazy complex pointer juggling, I wrote a "getter" that'd actually build a vector of pointers to the array of objects, and return that- i.e.:

```
std::vector<CItem*> items;
items.reserve( childCount );
for( size_t i = 0; i < childCount; ++i ) {
    items.emplace_back(children + i);
    }
return items;
```

...it actually did some work, rather than just return the value of the internal representation.

UIforETW/UIforETWDlg.cpp
ariccio: Whoops!

UIforETW/ChildProcess.cpp
ariccio: > Just a thought.

It's a good one. I agree. We should probably also have a macro that `ASSERT(false)`s on conditions that should never be reached, e.g. `UIETW_NOTREACHED()` so that we're explicit.

As a side note: you mentioned (somewhere else) reentrancy issues with `assert`, and I can easily guess/imagine what they are (calling `abort()`, and then not actually aborting seems bad, doesn't it?), but am not *exactly* sure what you were referring to.

Can you elaborate?

UIforETW/Settings.cpp
randomascii: > On the other hand, what would "encapsulation" look like here?

A getHwnd() function and no way to set it. i.e.; getter, but no setter. Preventing uncontrolled changes gives some wonderful guarantees.

UIforETW/ChildProcess.cpp
randomascii: Some versions of assert bring up a dialog box to let the user choose what to do. This can be quite brilliant, but if the dialog box is naively invoked on the current thread when it will run a message pump on the current thread, which affects behavior and may even allow execution to flow back to the assert location (i.e.; if you have an assert in WM_PAINT or WM_TIMER).

The fix is to open the dialog on an assert-specific thread (which is a bit of engineering work) or not open a dialog.

Anyway, I'm not convinced by my own UIETWASSERT argument since it is a bit ugly and non-standard and solves a *future* problem that may never occur. Therefore it probably counts as over-engineering.

UIforETW/DirectoryMonitor.cpp
randomascii: Yeah... but still, ugly. Especially the "_In_ DWORD ThreadId" which conveys zero information.

randomascii:
Let me know when it's ready and I'll merge. Ditto for the other one.

UIforETW/DirectoryMonitor.cpp
ariccio: Yeah, `_In_` (which GitHub "helpfully" formats if not enclosed in backticks) is kinda useless - it'd be nice if it did something like imply `_Const_`, or otherwise check access, but *nope* it doesn't.

I'm guessing that it (specifically doesn't) is an artifact of Microsoft using `SALInfer` on their entire codebase, which I can imagine producing 1,000,000+ false positives.

UIforETW/UIforETWDlg.cpp
```diff
@@ -271,16 +271,16 @@ BOOL CUIforETWDlg::OnInitDialog()
 	// Add "About..." menu item to system menu.
 
 	// IDM_ABOUTBOX must be in the system command range.
-	ASSERT((IDM_ABOUTBOX & 0xFFF0) == IDM_ABOUTBOX);
```
ariccio: Note to self: these should be `static_assert`s

ariccio:
I made the changes that (I think) you wanted. 

> I also added two `_Pre_satisfies_(nCode == HC_ACTION)` where we `ATLASSERT`ed function preconditions.

> Yeah, UIETWASSERT is a bit of overengineering, but it doesn't really detract from readability (custom assert macros are pretty common), and it opens up options for later.

Did you want me to remove `m_hWnd`?

UIforETW/DirectoryMonitor.cpp
randomascii: Well, _In_ isn't always useless -- if you pass "char*" the it provides some information, although ideally the lack of 'const' would convey that. _InOut_ is more useful. But _In_ is completely meaningless on a type such as DWORD, where it can't even meaningfully/usefully imply _Const_.

UIforETW/UIforETWDlg.cpp
randomascii: Good point.

UIforETW/UIforETWDlg.cpp
```diff
@@ -295,7 +294,7 @@ BOOL CUIforETWDlg::OnInitDialog()
 	}
 
 	wchar_t* windowsDir = nullptr;
-	VERIFY(SUCCEEDED(SHGetKnownFolderPath(FOLDERID_Windows, 0, NULL, &windowsDir)));
+	ATLVERIFY(SUCCEEDED(SHGetKnownFolderPath(FOLDERID_Windows, 0, NULL, &windowsDir)));
```
randomascii: If we made this a UIETWVERIFY we could resolve the /analyze warnings on these functions. Later.

randomascii:
I've come around to m_hWnd. I still think it is abhorrent that it is
accessible directly, but I can see the advantages of referencing m_hWnd
over *this, even though GetHwnd() would be better than either.

On Sun, May 10, 2015 at 11:17 PM, Alexander Riccio <notifications@github.com
> wrote:

> I made the changes that (I think) you wanted.
>
> I also added two _Pre_satisfies_(nCode == HC_ACTION) where we ATLASSERTed
> function preconditions.
>
> Yeah, UIETWASSERT is a bit of overengineering, but it doesn't really
> detract from readability (custom assert macros are pretty common), and it
> opens up options for later.
>
> Did you want me to remove m_hWnd?
>
> —
> <https://github.com/google/UIforETW/pull/25#issuecomment-100785263>.
>



-- 
Bruce Dawson

randomascii:
I just did a debug /analyze build and I got a lot of warnings which seemed to be related to this change. Most of them are on calls to SmartEnableWindow. The warnings happen on 32-bit and 64-bit.

c:\github\uiforetw\uiforetw\settings.cpp(147): warning C6388: 'm_hWnd' might not be '0':  this does not adhere to the specification for the function 'SHGetSpecialFolderPathW'. 
c:\github\uiforetw\uiforetw\utility.h(43): warning C28220: Integer expression expected for annotation 'SAL_satisfies', parameter 1, found '__formal(0,Win)'.
c:\github\uiforetw\uiforetw\utility.cpp(209): warning C28220: [Warning applies to SmartEnableWindow at c:\github\uiforetw\uiforetw\utility.h(43)]Integer expression expected for annotation 'SAL_satisfies', parameter 1, found '__formal(0,Win)'.
c:\github\uiforetw\uiforetw\utility.cpp(209): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(358): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(384): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(553): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(554): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(555): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(556): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(558): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(559): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(560): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(1200): warning C28020: The expression '0' is not true at this call.
c:\github\uiforetw\uiforetw\uiforetwdlg.cpp(1208): warning C28020: The expression '0' is not true at this call.

ariccio:
Strangely, changing `_Pre_satisfies_(Win)` to `_Pre_satisfies_(Win != NULL)` fixes those.

I'll submit a patch with that change in a minute.

randomascii:
/analyze has a long history of not handling boolean expressions robustly.
I'm glad that fix works.

Does it find bugs?

On Tue, May 12, 2015 at 4:23 PM, Alexander Riccio <notifications@github.com>
wrote:

> Strangely, changing _Pre_satisfies_(Win) to _Pre_satisfies_(Win != NULL)
> fixes those.
>
> I'll submit a patch with that change in a minute.
>
> —
> <https://github.com/google/UIforETW/pull/25#issuecomment-101458324>.
>



-- 
Bruce Dawson

ariccio:
[*maybe*](https://github.com/google/UIforETW/pull/29#issue-75777768)

I'm not sure if that's a bug or not, yet.

ariccio:
Sidenote:
MFC really goes overboard on using macro expansion to add a layer of "abstraction" to message maps:
`ON_BN_CLICKED(IDC_COPYSTARTUPPROFILE, &CSettings::OnBnClickedCopystartupprofile)` is expanded to `ON_CONTROL(BN_CLICKED, IDC_COPYSTARTUPPROFILE, &CSettings::OnBnClickedCopystartupprofile)`, which is further expanded to `{ WM_COMMAND, (WORD)BN_CLICKED, (WORD)id, (WORD)IDC_COPYSTARTUPPROFILE, AfxSigCmd_v, (static_cast< AFX_PMSG > (&CSettings::OnBnClickedCopystartupprofile)) }`

...which makes it *really fun* to try to track the usage/initialization of `m_hWnd`

randomascii:
Spoiler alert: it's set when the underlying window is created.

On Tue, May 12, 2015 at 4:40 PM, Alexander Riccio <notifications@github.com>
wrote:

> Sidenote:
> MFC really goes overboard on using macro expansion to add a layer of
> "abstraction" to message maps:
> ON_BN_CLICKED(IDC_COPYSTARTUPPROFILE,
> &CSettings::OnBnClickedCopystartupprofile) is expanded to ON_CONTROL(BN_CLICKED,
> IDC_COPYSTARTUPPROFILE, &CSettings::OnBnClickedCopystartupprofile), which
> is further expanded to { WM_COMMAND, (WORD)BN_CLICKED, (WORD)id,
> (WORD)IDC_COPYSTARTUPPROFILE, AfxSigCmd_v, (static_cast< AFX_PMSG >
> (&CSettings::OnBnClickedCopystartupprofile)) }
>
> ...which makes it *really fun* to try to track the usage/initialization
> of m_hWnd
>
> —
> <https://github.com/google/UIforETW/pull/25#issuecomment-101460723>.
>



-- 
Bruce Dawson

ariccio:
> Spoiler alert: it's set when the underlying window is created.

I figured that, but I'd like to make sure that `MFC`/`CWnd` isn't doing anything wacky while we're not looking.

The fact that analyze is saying that it might be zero, combined with the fact that `CWnd` provides the (mostly useless) `GetSafeHwnd` member function makes me suspicious.

I'd forgotten about `GetSafeHwnd`, because there are really an *absurd* number of member functions in `CWnd`.

randomascii:
Never make the mistake of thinking that /analyze's warnings are reasonable.

Sometimes they are bizarrely accurate but, unfortunately, if a skilled
developer cannot find the bug quickly then it is usually because /analyze
is insane.

Knowing it's patterns for brilliance versus insanity helps, and I hate to
counsel people to ignore it, but sometimes you have to. Sigh...

On Tue, May 12, 2015 at 4:51 PM, Alexander Riccio <notifications@github.com>
wrote:

> Spoiler alert: it's set when the underlying window is created.
>
> I figured that, but I'd like to make sure that MFC/CWnd isn't doing
> anything wacky while we're not looking.
>
> The fact that analyze is saying that it might be zero, combined with the
> fact that CWnd provides the (mostly useless) GetSafeHwnd member function
> makes me suspicious.
>
> I'd forgotten about GetSafeHwnd, because there are really an *absurd*
> number of member functions in CWnd.
>
> —
> <https://github.com/google/UIforETW/pull/25#issuecomment-101462020>.
>



-- 
Bruce Dawson

PR #27, opened 2015-05-07
add reserve, swap push_back with emplace_back

ariccio:
In places where we'd fill a `std::vector` inside a for loop, I added a call to `reserve` to preallocate the necessary space.

I replaced `push_back` with `emplace_back`, saving silly copy constructions.

In ranged-for loops, I changed `auto` to `auto&` to avoid copies, and added const wherever the items are not modified.

### DIFF
--- a/UIforETW/UIforETWDlg.cpp
+++ b/UIforETW/UIforETWDlg.cpp
@@ -1646,7 +1646,7 @@ void CUIforETWDlg::PreprocessTrace(const std::wstring& traceFilename)
 		child.Run(bShowCommands_, L"xperf.exe" + args);
 		std::wstring output = child.GetOutput();
 		std::map<std::wstring, std::vector<DWORD>> pidsByType;
-		for (const auto line : split(output, '\n'))
+		for (const auto& line : split(output, '\n'))
 		{
 			if (wcsstr(line.c_str(), L"chrome.exe"))
 			{
--- a/UIforETW/UIforETWDlg.cpp
+++ b/UIforETW/UIforETWDlg.cpp
@@ -1388,7 +1388,7 @@ void CUIforETWDlg::OnContextMenu(CWnd* pWnd, CPoint point)
 				ID_TRACES_TRACEPATHTOCLIPBOARD,
 			};
 
-			for (auto id : disableList)
+			for (const auto& id : disableList)
 				pContextMenu->EnableMenuItem(id, MF_BYCOMMAND | MF_GRAYED);
 		}
 
@@ -1569,7 +1569,7 @@ void CUIforETWDlg::CompressAllTraces() const
 	int64_t finalTotalSize = 0;
 	int notCompressedCount = 0;
 	int compressedCount = 0;
-	for (auto traceName : traces_)
+	for (const auto& traceName : traces_)
 	{
 		auto result = CompressTrace(GetTraceDir() + traceName + L".etl");
 		if (result.first == result.second)
@@ -1646,7 +1646,7 @@ void CUIforETWDlg::PreprocessTrace(const std::wstring& traceFilename)
 		child.Run(bShowCommands_, L"xperf.exe" + args);
 		std::wstring output = child.GetOutput();
 		std::map<std::wstring, std::vector<DWORD>> pidsByType;
-		for (auto& line : split(output, '\n'))
+		for (const auto line : split(output, '\n'))
 		{
 			if (wcsstr(line.c_str(), L"chrome.exe"))
 			{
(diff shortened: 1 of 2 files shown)


UIforETW/UIforETWDlg.cpp
```diff
@@ -1646,7 +1646,7 @@ void CUIforETWDlg::PreprocessTrace(const std::wstring& traceFilename)
 		child.Run(bShowCommands_, L"xperf.exe" + args);
 		std::wstring output = child.GetOutput();
 		std::map<std::wstring, std::vector<DWORD>> pidsByType;
-		for (auto& line : split(output, '\n'))
+		for (const auto line : split(output, '\n'))
```
randomascii: This line gained a const but lost a reference. It should still be a reference.

UIforETW/UIforETWDlg.cpp
```diff
@@ -1784,7 +1785,7 @@ void CUIforETWDlg::FinishTraceRename()
 				failedSource = oldName;
 				break;
 			}
-			renamed.push_back(std::pair<std::wstring, std::wstring>(oldName, newName));
+			renamed.emplace_back(std::make_pair(oldName, newName));
```
randomascii: Nice. I didn't know about make_pair.

UIforETW/Utility.cpp
```diff
@@ -274,22 +275,31 @@ std::wstring StripExtensionFromPath(const std::wstring& path)
...
-	paths.push_back(path);
+	paths.emplace_back(path);
 	return DeleteFiles(hwnd, paths);
 }
 
 int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
 {
+	ATLASSERT(paths.size() > 0);
+
+	size_t totalLength = 1;
```
randomascii: Not worth it. Yes, reserve is a good idea, but in this program it will *never* make a noticeable difference (vector is still O(N) even without reserve). So it is definitely not worth a loop to precalculate the required size.

I think that in UIforETW calling reserve() is mostly pretentious, and here it is definitely excessive. Less code is important also.

randomascii:
This is mostly an excellent change - just one bug, and one area that is too much.

UIforETW/Utility.cpp
ariccio: > in UIforETW calling reserve() is mostly pretentious

[...]

>  here it is definitely excessive

The defense pleads guilty to all charges.

UIforETW/UIforETWDlg.cpp
ariccio: It took me a while to figure out why it should be a reference - it *looks* a bit like we're assigning a return value to a reference, maybe it'd be better to store the result of split in a variable, and then iterate over it? 

i.e.:

```
auto lines = split(output, '\n');
for (const auto& line : lines)
```

randomascii: No need for a temporary variable. The ideal code is:

    for (const auto& line : split(output, '\n'))

Adding a 'lines' local variable is unnecessary, and would make the result of the split() function live longer than necessary. I think you're confusing the result of split() and the 'line' variable. The compiler is required to hang on to the return value of the split() function until containing expression (the for loop) completes. If it doesn't then it doesn't matter what the type of the 'line' variable is because it would be iterating through a destroyed vector. Therefore 'line' should be a reference because that is more efficient.

That is, making 'line' not be a reference wouldn't solve the problem you are concerned with, and it isn't a problem anyway.

I also like that the line of code ends up looking pythonic when we iterate over the return value directly. When C++11 lets code look more like Python I think that is a good thing.

ariccio: > I think you're confusing the result of split() and the 'line' variable. 

That's exactly what I was doing. I still find it a bit surprising, but that's fair enough.


> I also like that the line of code ends up looking pythonic when we iterate over the return value directly. When C++11 lets code look more like Python I think that is a good thing.

I agree entirely. Well, *almost* entirely. I really love `Python`, but it has some very strange scoping rules. [Loop variables leak into the enclosing scopes](http://eli.thegreenplace.net/2015/the-scope-of-index-variables-in-pythons-for-loops), which is *crazy* from a `C` state of mind.

ariccio:
I made the requested change (sorry for the delay!), although I didn't remove the length count & reserve  loop, did you want me to?

ariccio:
Note that I picked up two new code analysis warnings about a conditional (`Is64BitWindows() && sizeof(void*) != 8`), the first of which:
(settings.cpp 85)
```
C6240	Logical-AND with non-zero constant should be bitwise-AND	(<expression> && <non-zero constant>) always evaluates to the result of <expression>.  Did you intend to use the bitwise-and operator?	
```
The second is the exact same message, somewhere else.

randomascii:
If you could remove the length count and reserve loop that would great, then I can land it.

Bummer about the C6240 warning. It is sometimes useful but definitely has false positives such as this one.

randomascii:
I'll fix the code analysis warnings.

ariccio:
All clear!

PR #31, opened 2015-05-13
add more aggressive optimizations for release build

ariccio:
None of these have any noticeable effect (at the moment), but they're all a good idea.

[Global data optimization is pretty neat.](http://blogs.msdn.com/b/vcblog/archive/2013/09/11/introducing-gw-compiler-switch.aspx)

### DIFF
--- a/UIforETW/UIforETW.vcxproj
+++ b/UIforETW/UIforETW.vcxproj
@@ -178,7 +178,7 @@
       <TreatWarningAsError>true</TreatWarningAsError>
       <AdditionalIncludeDirectories>..\include</AdditionalIncludeDirectories>
       <EnablePREfast>false</EnablePREfast>
-      <AdditionalOptions> /Qvec-report:2 /Qpar-report:2 /Gw</AdditionalOptions>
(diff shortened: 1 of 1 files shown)


randomascii:
/Gw is cool (I enabled it for Chrome and it shrunk the binaries nicely) but the options that give reports on failed optimizations -- that's just noise, like this:

1>  --- Analyzing function: void __cdecl ATL::`dynamic atexit destructor for '_AtlWinModule''(void)
1>  c:\program files (x86)\microsoft visual studio 12.0\vc\atlmfc\include\atlsimpcoll.h(212) : info C5002: loop not vectorized due to reason '501'
1>  c:\program files (x86)\microsoft visual studio 12.0\vc\atlmfc\include\atlsimpcoll.h(212) : info C5012: loop not parallelized due to reason '501'

I don't *mind* more performance, but this project doesn't actually *need* it, so the build noise can't really be justified.

ariccio:
> so the build noise can't really be justified.

Fair enough. I cut 'em out.

PR #33, opened 2015-05-14
ETWProviders cleanup

ariccio:
Contrary to the name of this branch, I didn't actually investigate the Working Set monitoring performance issues. I did a bunch of cleanup instead.

The cleanup involved:
- Increasing warning level from 3 to 4
- Enabling Code Analysis for all builds (build time is still very fast)
- Adding annotations
- Converting private & unimplemented copy constructors to `delete`d functions
- etc...

There's one `/analyze` warning that I haven't yet figured out, but it's not a critical issue.

I have NOT included the recompiled `.lib`/`.dll` in this PR, as I figured that you might want to do that.

### DIFF
--- a/ETWProviders/etwprof.cpp
+++ b/ETWProviders/etwprof.cpp
@@ -137,7 +137,7 @@ ULONG EVNTAPI EventWrite( _In_ REGHANDLE RegHandle, _In_ PCEVENT_DESCRIPTOR Even
 }
 
 // Redirector function for EventUnregister. Called by macros in ETWProviderGenerated.h
-ULONG EVNTAPI EventUnregister( REGHANDLE RegHandle )
+ULONG EVNTAPI EventUnregister( _In_ REGHANDLE RegHandle )
 {
 	if ( g_ETWRegister.m_pEventUnregister )
 		return g_ETWRegister.m_pEventUnregister( RegHandle );
--- a/ETWProviders/etwprof.cpp
+++ b/ETWProviders/etwprof.cpp
@@ -120,6 +120,7 @@ public:
 } g_ETWRegister;
 
 // Redirector function for EventRegister. Called by macros in ETWProviderGenerated.h
+_Success_(return == ERROR_SUCCESS)
 ULONG EVNTAPI EventRegister( _In_ LPCGUID ProviderId, _In_opt_ PENABLECALLBACK EnableCallback, _In_opt_ PVOID CallbackContext, _Out_ PREGHANDLE RegHandle )
 {
 	if ( g_ETWRegister.m_pEventRegister )
@@ -129,6 +130,7 @@ ULONG EVNTAPI EventRegister( _In_ LPCGUID ProviderId, _In_opt_ PENABLECALLBACK E
 }
 
 // Redirector function for EventWrite. Called by macros in ETWProviderGenerated.h
+_Success_(return == ERROR_SUCCESS)
 ULONG EVNTAPI EventWrite( _In_ REGHANDLE RegHandle, _In_ PCEVENT_DESCRIPTOR EventDescriptor, _In_ ULONG UserDataCount, _In_reads_opt_(UserDataCount) PEVENT_DATA_DESCRIPTOR UserData )
 {
 	if ( g_ETWRegister.m_pEventWrite )
@@ -137,6 +139,7 @@ ULONG EVNTAPI EventWrite( _In_ REGHANDLE RegHandle, _In_ PCEVENT_DESCRIPTOR Even
 }
 
 // Redirector function for EventUnregister. Called by macros in ETWProviderGenerated.h
+_Success_(return == ERROR_SUCCESS)
 ULONG EVNTAPI EventUnregister( _In_ REGHANDLE RegHandle )
 {
 	if ( g_ETWRegister.m_pEventUnregister )
--- a/ETWProviders/etwprof.cpp
+++ b/ETWProviders/etwprof.cpp
@@ -165,31 +165,31 @@ static float QPCToMS( int64 nDelta )
 
 // Public functions for emitting ETW events.
 
-void ETWMark( const char *pMessage )
+void ETWMark( _In_z_ PCSTR pMessage )
 {
 	EventWriteMark( pMessage );
 }
 
-void ETWMark1I(const char* pMessage, int data1)
+void ETWMark1I(_In_z_ PCSTR pMessage, int data1)
 {
 	EventWriteMark1I( pMessage, data1 );
 }
 
-void ETWMark2I(const char* pMessage, int data1, int data2)
+void ETWMark2I(_In_z_ PCSTR pMessage, int data1, int data2)
 {
 	EventWriteMark2I( pMessage, data1, data2 );
 }
 
-void ETWMark1F(const char* pMessage, float data1)
+void ETWMark1F(_In_z_ PCSTR pMessage, float data1)
 {
 	EventWriteMark1F( pMessage, data1 );
 }
-void ETWMark2F(const char* pMessage, float data1, float data2)
+void ETWMark2F(_In_z_ PCSTR pMessage, float data1, float data2)
 {
 	EventWriteMark2F( pMessage, data1, data2 );
 }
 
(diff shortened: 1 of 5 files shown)


UIforETW/WorkingSet.cpp
```diff
@@ -68,13 +73,16 @@ void CWorkingSetMonitor::SampleWorkingSets()
...
+			{
 				match = true;
+			}
 		}
 		if (match)
 		{
 			DWORD pid = peInfo.th32ProcessID;
 			// Get a handle to the process.
-			HANDLE hProcess = OpenProcess(PROCESS_QUERY_INFORMATION |
+			HANDLE hProcess =
```
randomascii: Beware of reformatting the code for minor taste reasons...

UIforETW/WorkingSet.cpp
```diff
@@ -48,8 +48,13 @@ void CWorkingSetMonitor::SampleWorkingSets()
...
 	ULONG_PTR numEntries = 100000;
-	std::vector<char> buffer(sizeof(PSAPI_WORKING_SET_INFORMATION) + numEntries * sizeof(PSAPI_WORKING_SET_BLOCK));
-	PSAPI_WORKING_SET_INFORMATION* pwsBuffer = reinterpret_cast<PSAPI_WORKING_SET_INFORMATION*>(&buffer[0]);
+
+	const rsize_t bufferSizeNeeded =
+		sizeof(PSAPI_WORKING_SET_INFORMATION) +
+		(numEntries * sizeof(PSAPI_WORKING_SET_BLOCK));
+
+	std::vector<char> buffer(bufferSizeNeeded);
+	PSAPI_WORKING_SET_INFORMATION* pwsBuffer = reinterpret_cast<PSAPI_WORKING_SET_INFORMATION*>(buffer.data());
```
randomascii: buffer.data() is not clearly better than &buffer[0]. I've always preferred &buffer[0] because it seems very clear and because it works with arrays and vector.

include/ETWProviders/etwprof.h
```diff
@@ -39,32 +39,32 @@ const int kFlagDoubleClick = 100;
...
-PLATFORM_INTERFACE void ETWMark2F(const char* pMessage, float data1, float data2);
+PLATFORM_INTERFACE void ETWMark1I(_In_z_ PCSTR pMessage, int data1);
+PLATFORM_INTERFACE void ETWMark2I(_In_z_ PCSTR pMessage, int data1, int data2);
+PLATFORM_INTERFACE void ETWMark1F(_In_z_ PCSTR pMessage, float data1);
+PLATFORM_INTERFACE void ETWMark2F(_In_z_ PCSTR pMessage, float data1, float data2);
 
 // _Printf_format_string_ is used by /analyze
-PLATFORM_INTERFACE void ETWMarkPrintf( _Printf_format_string_ const char *pMessage, ... );
-PLATFORM_INTERFACE void ETWWorkerMarkPrintf( _Printf_format_string_ const char *pMessage, ... );
+PLATFORM_INTERFACE void ETWMarkPrintf( _Printf_format_string_ _In_z_ PCSTR pMessage, ... );
```
randomascii: If /analyze needs both _Printf_format_string *and* _In_z_ then it is being very stupid. One should imply the other. I'm not saying it does. I'm just saying that it *should*. Do you know if it does? No point adding _In_z_ if it adds no value. That would be like adding _In_ to an int64.

ETWProviders/etwprof.cpp
```diff
@@ -120,28 +120,35 @@ class CETWRegister
 } g_ETWRegister;
 
 // Redirector function for EventRegister. Called by macros in ETWProviderGenerated.h
-ULONG EVNTAPI EventRegister( LPCGUID ProviderId, PENABLECALLBACK EnableCallback, PVOID CallbackContext, PREGHANDLE RegHandle )
+#pragma warning(suppress: 28253)
```
randomascii: It is a good policy to *never* put in a warning suppression #pragma without explaining why. The simplest way to do that is with a comment that includes the warning text, 'cause who is going to remember what 28253 is?

ETWProviders/etwprof.cpp
```diff
@@ -120,28 +120,35 @@ class CETWRegister
 } g_ETWRegister;
 
 // Redirector function for EventRegister. Called by macros in ETWProviderGenerated.h
-ULONG EVNTAPI EventRegister( LPCGUID ProviderId, PENABLECALLBACK EnableCallback, PVOID CallbackContext, PREGHANDLE RegHandle )
+#pragma warning(suppress: 28253)
+_Success_(return == ERROR_SUCCESS)
```
randomascii: This annotation doesn't seem to be working, and isn't present in Microsoft's definition of this function.

Why add the annotation if it doesn't seem to work (i.e.; warning C6101).

UIforETW/WorkingSet.cpp
ariccio: 80 column limit?

ETWProviders/etwprof.cpp
ariccio: Good point......I was annoyed enough with the fact that Microsoft didn't properly annotate their own function - moreso than usual because we are redirecting - that I wasn't thinking. I should fix this tomorrow?

randomascii: I fixed already, and got the warning as well.

include/ETWProviders/etwprof.h
ariccio: > Do you know if it does?

I have no evidence to support any conclusion, but I don't think that it does.

> If /analyze needs both `_Printf_format_string_ and `_In_z_ then it is being very stupid. One should imply the other. I'm not saying it does. I'm just saying that it *should*.

I'm *not sure* that it *should*. I wouldn't be surprised if there was some absurd use case where a `_Printf_format_string_` could also be `_Out_writes_z_`....I'd never advocate writing code like that, but someone else *might*.

It [seems that other users of `_Printf_format_string_` also combine `_In_z_` with `_Printf_format_string_`](https://www.osronline.com/showthread.cfm?link=255907).

randomascii: Nah, it's not possible. The contact for a _Printf_format_string_ is pretty
straightforward -- it behaves like printf. That means it is a
null-terminated string with % specifiers that match the arguments. If that
isn't true then it's not a _Printf_format_string_ but something similar.

I believe that _In_z_ has been combined with _Printf_format_string_
elsewhere, and probably by Microsoft too, but that just means they designed
it poorly, IMHO.

On Fri, May 15, 2015 at 7:04 PM, Alexander Riccio <notifications@github.com>
wrote:

> In include/ETWProviders/etwprof.h
> <https://github.com/google/UIforETW/pull/33#discussion_r30457347>:
>
> >
> >  // _Printf_format_string_ is used by /analyze
> > -PLATFORM_INTERFACE void ETWMarkPrintf( _Printf_format_string_ const char *pMessage, ... );
> > -PLATFORM_INTERFACE void ETWWorkerMarkPrintf( _Printf_format_string_ const char *pMessage, ... );
> > +PLATFORM_INTERFACE void ETWMarkPrintf( _Printf_format_string_ _In_z_ PCSTR pMessage, ... );
>
>  Do you know if it does?
>
> I have no evidence to support any conclusion, but I don't think that it
> does.
>
> If /analyze needs both _Printf_format_string_ and*In_z* then it is being
> very stupid. One should imply the other. I'm not saying it does. I'm just
> saying that it *should*.
>
> I'm *not sure* that it *should*. I wouldn't be surprised if there was
> some absurd use case where a _Printf_format_string_ could also be
> _Out_writes_z_....I'd never advocate writing code like that, but someone
> else *might*.
>
> It seems that other users of _Printf_format_string_ also combine _In_z_
> with _Printf_format_string_
> <https://www.osronline.com/showthread.cfm?link=255907>.
>
> —
> <https://github.com/google/UIforETW/pull/33/files#r30457347>.
>



-- 
Bruce Dawson

ariccio: > but that just means they designed it poorly, IMHO.

That wouldn't surprise me. SAL has a lot of great applications, and lots of potential, but given it's evolution as an internal Microsoft tool, I doubt they got enough useful feedback.

PR #39, opened 2015-06-22
fix python 2/3 dict.keys method inconsistency

ariccio:
[The behavior of dict.keys changed in Python 3](http://blog.labix.org/2008/06/27/watch-out-for-listdictkeys-in-python-3). There's [a version-agnostic way to do this](http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items).

Without this, Python crashes (i.e. a `Traceback (most recent call last):` "crash").


randomascii:
Cool.

If the scripts are now Python version agnostic do you want to remove the pytwoseven check from FindPython()?

ariccio:
*Maybe*. I'll review the (Python) code carefully before saying that.

randomascii:
No hurry/no worry.

Issue #2, opened 2015-04-12
An installer to install the MFC DLLs is needed


randomascii:
In order to keep the UIforETW binary small so that it can be occasionally checked in to github without bloating the repo it is built to use the DLL versions of the C run-time and MFC. However this means that it doesn't work on machines that don't have these DLLs installed. An installer that installs these DLLs could be checked in once and then used with multiple versions of UIforETW.

randomascii:
The use of github releases and static linking means that this is no longer necessary. The releases will run on any machine and this is a simpler to use (and therefore better) mechanism than a separate installer for the MFC and CRT DLLs.

PR #40, opened 2015-06-24
add comments to clarify obscure xperf args

ariccio:
Hopefully, self-descriptive.

### DIFF
--- a/bin/StripChromeSymbols.py
+++ b/bin/StripChromeSymbols.py
@@ -100,6 +100,10 @@ def main():
   # to stop xperf from using -- rename them from .pdb to .pdbx
   local_symbol_files = []
 
+  #-tle = tolerate lost events
+  #-tti = tolerate time ivnersions
+  #-a symcache = show image and symbol identification
+  #-dbgid = show symbol identification information
   command = 'xperf -i "%s" -tle -tti -a symcache -dbgid' % tracename
   print("> %s" % command)
   found_uncached = False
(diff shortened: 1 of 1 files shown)


randomascii:
I experimented with a few merge techniques:
1) The recommended manual merge technique with --no-ff causes weird reordering stuff that I don't like.
2) Removing --no-ff creates a more straightforward history. In this case it adds your commit (author == you) and a merge commit (author == me).
3) Using --squash leads to the simplest result - a single commit - but it means that the author is listed as me (with your information in the description).

Thoughts/preferences? I'm going to use --squash for complex changes, but I could use option #2 for simpler cases. It depends, to some extent, on how much you want yourself listed as 'author'.

hymerman:
Option 2 is the normal way; generally it's best to see the commit history complete with authors - squashing complex work into single large commits is unusual and makes it harder to see exactly what was done, and harder to revert parts selectively.

If there are lots of useless commits (say, of 'work in progress' commits, or lots of repeated merges from upstream), it's best for contributors to simplify this by rebasing/squashing to remove those commits, prior to submitting pull requests. If that doesn't happen it's ok to ask them to go back and massage their branch history a little and resubmit the pull request.

randomascii:
Sounds good. I'll save squash for emergency situations and ask contributors to use it if it seems necessary. If omitting --no-ff avoids the weird history problems then I'll be happy.

hymerman:
Cool! I should have said; --no-ff basically exists explicitly to make
history a bit weird - it'll force a merge commit even when one isn't
needed. This way you *could* view the DAG of history as one line containing
your commits (including merge commits), which contributors branch from and
get merged into. Otherwise you'll get little runs where there are commits
by people other than yourself. I find it a bit pointless but some like it.

PR #41, opened 2015-07-19
Fix `LocalAlloc`/`pdidd` NULL deref warning

ariccio:
`/analyze` complained:

```
C28182 Dereferencing a copy of a null pointer 'pdidd' contains the same
NULL value as 'LocalAlloc()`128' did.
139 'pdidd' may be NULL
140 'pdidd' is dereferenced, but may still be NULL
```

So, I inserted a `NULL` check, and simply `std::terminate()` if
`LocalAlloc` fails, as it means we're probably OOM (unless there's an
***absurdly*** large `DevicePath` in `SP_DEVICE_INTERFACE_DETAIL_DATA`).

There were so many levels of indentation in
`CPowerStatusMonitor::SampleBatteryStat` that it was giving me a
headache, so I went ahead and 'unrolled' it. Is that ok?

Because I'm crazy about checking return values, I added two wrapper
functions to `Utility.h`/`Utility.cpp` (`FreeLocalAlloc` and
`CloseValidHandle`), as if the wrapped system calls ever fail, there's a
bug on our side of things.

As a drive-by fix, I also changed the `OUT` pseudo-annotated parameters
of `NtQueryTimerResolution` (`OUT` is defined to ` ` - blank text) to
`_Out_`, so that `/analyze` treats them as actual out parameters.

**PLEASE NOTE**:

There's another code analysis warning that is ***NOT*** correct

```
C6102		Using 'bytesNeeded' from failed function call at line '132'.
	128	'bytesNeeded' is not initialized
	135	Skip this branch, (assume 'getDetailResult' is false)
	140	Skip this branch, (assume 'GetLastError()!=122' is false)
	146	'bytesNeeded' is an Input to 'LocalAlloc' (declared at c:\program files (x86)\windows kits\8.1\include\um\winbase.h:1039)
	146	'bytesNeeded' is used, but may not have been initialized
```

... Microsoft just hasn't properly annotated `SetupDiGetDeviceInterfaceDetail`
to say that the 5th parameter is valid under certain failure conditions
(i.e. `_On_failure_( _When_( ( _Param_( 3 ) == 0 ) && ( _Param_( 4 ) == 0 ), _Post_valid_ ) )`...)

### DIFF
--- a/UIforETW/PowerStatus.cpp
+++ b/UIforETW/PowerStatus.cpp
@@ -108,125 +108,179 @@ void CPowerStatusMonitor::SampleBatteryStat()
 	for (int deviceNum = 0; deviceNum < maxBatteries; deviceNum++)
 	{
 		SP_DEVICE_INTERFACE_DATA did = { sizeof(did) };
-
-		if (SetupDiEnumDeviceInterfaces(hdev,
+		const BOOL enumDeviceInterfacesResult =
+			SetupDiEnumDeviceInterfaces(hdev,
 										0,
 										&GUID_DEVCLASS_BATTERY,
 										deviceNum,
-										&did))
+										&did);
+		if (!enumDeviceInterfacesResult)
 		{
-			DWORD bytesNeeded = 0;
-			SetupDiGetDeviceInterfaceDetail(hdev,
+			const DWORD lastErr = ::GetLastError( );
+			if (lastErr == ERROR_NO_MORE_ITEMS)
+			{
+				break; // Enumeration failed - perhaps we're out of items
+			}
+			continue;
+		}
+
+		DWORD bytesNeeded = 0;
+		const BOOL getDetailResult = SetupDiGetDeviceInterfaceDetail(hdev,
+										&did,
+										0,
+										0,
+										&bytesNeeded,
+										0);
+
+		if (getDetailResult)
+		{
+			throw std::logic_error("Expected SetupDiGetDeviceInterfaceDetail failure!");
+		}
+
+		if (GetLastError() != ERROR_INSUFFICIENT_BUFFER)
+		{
+			continue;
+		}
+
+
+		HLOCAL detailDataAlloc = LocalAlloc(LPTR, bytesNeeded);
+		if (detailDataAlloc == NULL)
+		{
+			std::terminate();
+		}
+
+		auto pdidd =
+			reinterpret_cast<PSP_DEVICE_INTERFACE_DETAIL_DATA const>(detailDataAlloc);
+		pdidd->cbSize = sizeof(*pdidd);
+		
+		const BOOL batteryFound = SetupDiGetDeviceInterfaceDetail(hdev,
 											&did,
-											0,
-											0,
+											pdidd,
+											bytesNeeded,
 											&bytesNeeded,
 											0);
-
-			if (GetLastError() == ERROR_INSUFFICIENT_BUFFER)
-			{
-				auto pdidd = (PSP_DEVICE_INTERFACE_DETAIL_DATA)LocalAlloc(LPTR, bytesNeeded);
-				pdidd->cbSize = sizeof(*pdidd);
-				if (SetupDiGetDeviceInterfaceDetail(hdev,
-													&did,
-													pdidd,
-													bytesNeeded,
-													&bytesNeeded,
-													0))
-				{
-					// Found a battery. Query it.
-					HANDLE hBattery = CreateFile(pdidd->DevicePath,
-									GENERIC_READ | GENERIC_WRITE,
-									FILE_SHARE_READ | FILE_SHARE_WRITE,
-									NULL,
-									OPEN_EXISTING,
-									FILE_ATTRIBUTE_NORMAL,
-									NULL);
-					if (hBattery != INVALID_HANDLE_VALUE)
-					{
-						BATTERY_QUERY_INFORMATION bqi = {};
-
-						DWORD dwWait = 0;
-						DWORD dwOut;
-						if (DeviceIoControl(hBattery,
-											IOCTL_BATTERY_QUERY_TAG,
-											&dwWait,
-											sizeof(dwWait),
-											&bqi.BatteryTag,
-											sizeof(bqi.BatteryTag),
-											&dwOut,
-											NULL)
-											&& bqi.BatteryTag)
-						{
-							// With the tag, you can query the battery info.
-							BATTERY_INFORMATION bi = {};
-							bqi.InformationLevel = BatteryInformation;
-
-							if (DeviceIoControl(hBattery,
-												IOCTL_BATTERY_QUERY_INFORMATION,
-												&bqi,
-												sizeof(bqi),
-												&bi,
-												sizeof(bi),
-												&dwOut,
-												NULL))
-							{
-								// Only non-UPS system batteries count
-								if (bi.Capabilities & BATTERY_SYSTEM_BATTERY)
-								{
-									// Query the battery status.
-									BATTERY_WAIT_STATUS bws = {};
-									bws.BatteryTag = bqi.BatteryTag;
-
-									BATTERY_STATUS bs;
-									if (DeviceIoControl(hBattery,
-														IOCTL_BATTERY_QUERY_STATUS,
-														&bws,
-														sizeof(bws),
-														&bs,
-														sizeof(bs),
-														&dwOut,
-														NULL))
-									{
-										char powerState[100];
-										powerState[0] = 0;
-										if (bs.PowerState & BATTERY_CHARGING)
-											strcat_s(powerState, "Charging");
-										if (bs.PowerState & BATTERY_DISCHARGING)
-											strcat_s(powerState, "Discharging");
-										if (bs.PowerState & BATTERY_POWER_ON_LINE)
-										{
-											if (powerState[0])
-												strcat_s(powerState, ", on AC power");
-											else
-												strcat_s(powerState, "On AC power");
-										}
-
-										float batPercentage = bs.Capacity * 100.f / bi.FullChargedCapacity;
-
-										char rate[100];
-										if (bs.Rate == BATTERY_UNKNOWN_RATE)
-											sprintf_s(rate, "Unknown rate");
-										else if (bi.Capabilities & BATTERY_CAPACITY_RELATIVE)
-											sprintf_s(rate, "%ld (unknown units)", bs.Rate);
-										else
-											sprintf_s(rate, "%1.3f watts", bs.Rate / 1000.0f);
-										ETWMarkBatteryStatus(powerState, batPercentage, rate);
-										//outputPrintf(L"%S, %1.3f%%, %S\n", powerState, batPercentage, rate);
-									}
-								}
-							}
-						}
-						CloseHandle(hBattery);
-					}
-				}
-				LocalFree(pdidd);
-			}
-		}
-		else  if (ERROR_NO_MORE_ITEMS == GetLastError())
+		if (!batteryFound)
 		{
-			break;  // Enumeration failed - perhaps we're out of items
+			FreeLocalAlloc(detailDataAlloc);//use `detailDataAlloc` to suppress
+			                                //analyze warning about using
+			                                //`pdidd` from failed
+			                                //`SetupDiGetDeviceInterfaceDetail`
+			                                //call...
+			continue;
 		}
+
+		// Found a battery. Query it.
+		HANDLE hBattery = CreateFile(pdidd->DevicePath,
+						GENERIC_READ | GENERIC_WRITE,
+						FILE_SHARE_READ | FILE_SHARE_WRITE,
+						NULL,
+						OPEN_EXISTING,
+						FILE_ATTRIBUTE_NORMAL,
+						NULL);
+
+		if (hBattery == INVALID_HANDLE_VALUE)
+		{
+			FreeLocalAlloc(pdidd);
+			continue;
+		}
+
+		BATTERY_QUERY_INFORMATION bqi = {};
+
+		DWORD dwWait = 0;
+		DWORD dwOut;
+
+		const BOOL tagQueried = DeviceIoControl(hBattery,
+												IOCTL_BATTERY_QUERY_TAG,
+												&dwWait,
+												sizeof(dwWait),
+												&bqi.BatteryTag,
+												sizeof(bqi.BatteryTag),
+												&dwOut,
+												NULL);
+
+		if (!(tagQueried && bqi.BatteryTag))
+		{
+			FreeLocalAlloc(pdidd);
+			CloseValidHandle(hBattery);
+			continue;
+		}
+
+		// With the tag, you can query the battery info.
+		BATTERY_INFORMATION bi = {};
+		bqi.InformationLevel = BatteryInformation;
+		
+		const BOOL infoQueried = DeviceIoControl(hBattery,
+												 IOCTL_BATTERY_QUERY_INFORMATION,
+												 &bqi,
+												 sizeof(bqi),
+												 &bi,
+												 sizeof(bi),
+												 &dwOut,
+												 NULL);
+		if (!infoQueried)
+		{
+			FreeLocalAlloc(pdidd);
+			CloseValidHandle(hBattery);
+			continue;
+		}
+
+		// Only non-UPS system batteries count
+		if (!(bi.Capabilities & BATTERY_SYSTEM_BATTERY))
+		{
+			FreeLocalAlloc(pdidd);
+			CloseValidHandle(hBattery);
+			continue;
+		}
+
+
+		// Query the battery status.
+		BATTERY_WAIT_STATUS bws = {};
+		bws.BatteryTag = bqi.BatteryTag;
+
+		BATTERY_STATUS bs;
+		
+		const BOOL statusQueried = DeviceIoControl(hBattery,
+												   IOCTL_BATTERY_QUERY_STATUS,
+												   &bws,
+												   sizeof(bws),
+												   &bs,
+												   sizeof(bs),
+												   &dwOut,
+												   NULL);
+		if (!statusQueried)
+		{
+			FreeLocalAlloc(pdidd);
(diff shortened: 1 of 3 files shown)


UIforETW/PowerStatus.cpp
```diff
@@ -108,125 +108,179 @@ void CPowerStatusMonitor::SampleBatteryStat()
...
+		HLOCAL detailDataAlloc = LocalAlloc(LPTR, bytesNeeded);
+		if (detailDataAlloc == NULL)
 		{
-			DWORD bytesNeeded = 0;
-			SetupDiGetDeviceInterfaceDetail(hdev,
+			std::terminate();
+		}
+
+		auto pdidd =
+			reinterpret_cast<PSP_DEVICE_INTERFACE_DETAIL_DATA const>(detailDataAlloc);
```
randomascii: Since this isn't Chrome we can just go out to ~90 columns and keep it readable.

UIforETW/PowerStatus.cpp
```diff
@@ -108,125 +108,179 @@ void CPowerStatusMonitor::SampleBatteryStat()
...
+			throw std::logic_error("Expected SetupDiGetDeviceInterfaceDetail failure!");
+		}
+
+		if (GetLastError() != ERROR_INSUFFICIENT_BUFFER)
+		{
+			continue;
+		}
+
+
+		HLOCAL detailDataAlloc = LocalAlloc(LPTR, bytesNeeded);
```
randomascii: If refactoring why not makes this a std::vector<BYTE>? That avoids the NULL check, the confusing FreeLocalAlloc calls, etc.

UIforETW/PowerStatus.cpp
```diff
@@ -108,125 +108,179 @@ void CPowerStatusMonitor::SampleBatteryStat()
...
+		char rate[100];
+		if (bs.Rate == BATTERY_UNKNOWN_RATE)
+			sprintf_s(rate, "Unknown rate");
+		else if (bi.Capabilities & BATTERY_CAPACITY_RELATIVE)
+			sprintf_s(rate, "%ld (unknown units)", bs.Rate);
+		else
+			sprintf_s(rate, "%1.3f watts", bs.Rate / 1000.0f);
+		ETWMarkBatteryStatus(powerState, batPercentage, rate);
+		//outputPrintf(L"%S, %1.3f%%, %S\n", powerState, batPercentage, rate);
+		CloseValidHandle(hBattery);
```
randomascii: Way too many calls to CloseValidHandle and FreeLocalAlloc(pdidd). That much copy/paste means that it is definitely for resource management classes. std::vector for the memory, and some handle wrapper class for the handle. That's nicer than adding wrapper functions anyway.

UIforETW/PowerStatus.cpp
```diff
@@ -237,9 +291,9 @@ typedef _Return_type_success_(return >= 0) LONG NTSTATUS;
 // NtQueryTimerResolution is not officially documented.
 const NTSTATUS STATUS_SUCCESS = 0;
 extern "C" NTSYSAPI NTSTATUS NTAPI NtQueryTimerResolution(
-	OUT PULONG minimumResolution,
-	OUT PULONG maximumResolution,
-	OUT PULONG currentResolution);
+	_Out_ PULONG minimumResolution,
```
randomascii: Sure.

UIforETW/Utility.cpp
```diff
@@ -633,3 +633,24 @@ void CopyStartupProfiles(const std::wstring& exeDir, bool force)
 		}
 	}
 }
+
+void CloseValidHandle( _Pre_valid_ _Post_ptr_invalid_ HANDLE handle )
```
randomascii: Helper classes > helper functions.

UIforETW/PowerStatus.cpp
```diff
@@ -108,125 +108,179 @@ void CPowerStatusMonitor::SampleBatteryStat()
...
+			FreeLocalAlloc(pdidd);
+			CloseValidHandle(hBattery);
+			continue;
+		}
+
+		// With the tag, you can query the battery info.
+		BATTERY_INFORMATION bi = {};
+		bqi.InformationLevel = BatteryInformation;
+		
+		const BOOL infoQueried = DeviceIoControl(hBattery,
```
randomascii: Every variable in scope increases complexity. Sometimes naming a result is necessary, or sometimes it increases understanding, but in most of these cases I don't think that is so. I think just "if (!DeviceIoControl(...)) continue;" would be better.

UIforETW/PowerStatus.cpp
```diff
@@ -108,125 +108,179 @@ void CPowerStatusMonitor::SampleBatteryStat()
...
 											&did,
-											0,
-											0,
+											pdidd,
+											bytesNeeded,
 											&bytesNeeded,
 											0);
+		if (!batteryFound)
+		{
+			FreeLocalAlloc(detailDataAlloc);//use `detailDataAlloc` to suppress
```
randomascii: I don't understand why detailDataAlloc avoids any warning. The only warning I saw was a NULL pointer warning and that is avoided by checking for NULL (as you did) or by using std::vector.

randomascii:
Thanks for looking into these. See my comments on the particulars.

There are also warnings on SetThreadName. On release builds there are also these two:
d:\src\uiforetw\uiforetw\uiforetwdlg.cpp(304): warning C6102: Using 'progFilesx86Dir' from failed function call at line '303'.
d:\src\uiforetw\uiforetw\uiforetwdlg.cpp(293): warning C6102: Using 'windowsDir' from failed function call at line '292'.

VERIFY isn't *quite* the right semantics since it does nothing in release builds.

I really wish Windows had more no-fail functions.

UIforETW/PowerStatus.cpp
ariccio: Ah? I figured that there was some reason why you chose `LocalAlloc` over some other memory allocation function....

randomascii: Heh - no. I started from the MSDN sample code at https://msdn.microsoft.com/en-us/library/windows/desktop/bb204769(v=vs.85).aspx and stopped refactoring once it worked, more or less. I probably should have switched to std::vector<> right then.

A quick check confirms that LocalAlloc has been mostly obsolete since Win32 came into existence 20+ years ago (https://msdn.microsoft.com/en-us/library/windows/desktop/aa366533%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396), so I'm not sure why MSDN samples use it. There are a few places that it is required (clipboard?) but they are rare and this does not appear to be one of them.

ariccio: > A quick check confirms that LocalAlloc has been mostly obsolete since Win32 came into existence 20+ years ago

Yeah, pretty much. Raymond Chen, as always, [has some insight into this odd relic of 16-bit Windows](http://blogs.msdn.com/b/oldnewthing/archive/2004/11/01/250610.aspx).

The fact that the MSDN sample (for an API that's less than 20 years old?) uses it, makes me a bit nervous about changing it - did I miss something in the documentation?

randomascii: All that means is that the doc writer had not upgraded their habits. Sloppy. LocalAlloc or GlobalAlloc or required for functions that take a memory handle or which specifically request funny memory types. If they aren't specified then you can assume that any old pointer will do.

UIforETW/PowerStatus.cpp
ariccio: Fair enough. In a way, I find function calls nested inside `if`-statement-conditions less readable, especially for parameter crazy functions like `DeviceIoControl`.

I'll change it when I'm at a computer.

UIforETW/PowerStatus.cpp
ariccio: I think you've spotted my inner C programmer. It's never quite died out :)

I'll change it.

UIforETW/Utility.cpp
```diff
@@ -633,3 +633,24 @@ void CopyStartupProfiles(const std::wstring& exeDir, bool force)
...
+
+void FreeLocalAlloc( _Post_ptr_invalid_ HLOCAL alloc )
+{
+	const HLOCAL nullIfFreed = ::LocalFree(alloc);
+	if (nullIfFreed != NULL)
+	{
+		std::terminate( );
+	}
+
+}
```
ariccio: Hrmph. Oops.

UIforETW/PowerStatus.cpp
ariccio: Is "sloppy" even the proper word? I think "insane" is probably more descriptive.

UIforETW/PowerStatus.cpp
```diff
@@ -108,125 +108,179 @@ void CPowerStatusMonitor::SampleBatteryStat()
...
+			                                //analyze warning about using
+			                                //`pdidd` from failed
+			                                //`SetupDiGetDeviceInterfaceDetail`
+			                                //call...
+			continue;
+		}
 
-			if (GetLastError() == ERROR_INSUFFICIENT_BUFFER)
```
ariccio: By the way, what was the intended purpose of this call? As best as I can tell from the [`SetupDiGetDeviceInterfaceDetail` documentation](https://msdn.microsoft.com/en-us/library/windows/hardware/ff551120.aspx), this should always be true when called to find the needed buffer size?

randomascii: As always it is unclear whether the return value is guaranteed or not, so we might as well keep the check. I just cribbed it from the sample code (my post popular excuse apparently).

UIforETW/PowerStatus.cpp
```diff
@@ -108,125 +108,179 @@ void CPowerStatusMonitor::SampleBatteryStat()
...
+		const BOOL getDetailResult = SetupDiGetDeviceInterfaceDetail(hdev,
+										&did,
+										0,
+										0,
+										&bytesNeeded,
+										0);
+
+		if (getDetailResult)
+		{
+			throw std::logic_error("Expected SetupDiGetDeviceInterfaceDetail failure!");
```
ariccio: Would you rather, here:

- a `std::terminate()`
- a `throw std::logic_error`
- a `UIETWASSERT`

randomascii: The trouble with throw is that I wouldn't be surprised if MFC actually catches it, thus silently suppressing the error. Yuck. std::terminate()? I don't know.

ariccio: > The trouble with throw is that I wouldn't be surprised if MFC actually catches it, thus silently suppressing the error. 

LOL I didn't think about that, but it's *totally* plausible.

I hate `std::terminate` too.... Maybe a `UIETWASSERT` and a loud `outputPrintf`?

randomascii: I think use whichever is most compact. The expectation is that this will never occur so I'd hate to devote multiple lines of code to it. That means either std::terminate, or UIETWASSERT, I think.

UIforETW/Utility.cpp
```diff
@@ -633,3 +633,24 @@ void CopyStartupProfiles(const std::wstring& exeDir, bool force)
...
+	const BOOL handleClosed = ::CloseHandle(handle);
+	if (handleClosed == 0)
+	{
+		std::terminate( );
+	}
+
+}
+
+
+void FreeLocalAlloc( _Post_ptr_invalid_ HLOCAL alloc )
```
ariccio: Hmm. Forgot to remove this.

ariccio:
So, I think I addressed everything. I threw together a few helper functions/objects, and pulled the string formatting code out into a separate function. They're all in an anonymous namespace, if that's ok?

If I have addressed everything, let me know, and I'll open a separate - squashed - Pull Request.

randomascii:
The _At_ annotations on the destructors seem... odd. They seem to be indicating that the hBattery member will be invalid after the destructor runs, but since the entire object is invalid after the destructor runs that doesn't seem to add a lot of value.

The helper classes should be marked as non-copyable, presumably by marking the copy constructor and assignment operator as non-copyable. Otherwise they become a vector for new bugs.

There are some style inconsistencies (parentheses on the same line on struct and namespace declarations) but I'm not sure how much I care about them.

Using std::vector certainly helps a lot.

Have you tested on a machine with a battery? I haven't, but I can.

Note that CloseValidHandle is only called from one place now. Is it really worth having it as a separate function? Perhaps better to just do the check at that one point - that avoids having the reader have to go look and see what this odd function actually does.

I'm not convinced that setting up deviceIFaceDetailDataByteSize in order to avoid one /analyze warning is a good tradeoff. /analyze should make code *better*, not messier. I would have just used #pragma warning(suppress). But, I don't care that much either way.

ariccio:
> The At annotations on the destructors seem... odd. They seem to be indicating that the hBattery member will be invalid after the destructor runs, but since the entire object is invalid after the destructor runs that doesn't seem to add a lot of value.

Yeah, they're strange, redundant, preemptive, and probably unnecessary.

I don't yet have a copy of VS2015, so I can't check if it's been fixed, but otherwise `/analyze` has trouble understanding has trouble understanding what/how resources are released in destructors. The `_Post_ptr_invalid_` is that handle-invalidation hack that I described elsewhere, to detect handle misuse. Otherwise, `/analyze` has no way of telling that it's not just the memory that's *backing* the handle that's invalid after the destructor runs, but *the resource that the handle refers to* which is invalid. The `_ptr_` part, which is why I refer to this as a hack, works (*I think*) because a `HANDLE` *is just a* `typedef`'d `void*`.

(*to be continued *)

ariccio:
> The helper classes should be marked as non-copyable, presumably by marking the copy constructor and assignment operator as non-copyable. Otherwise they become a vector for new bugs.

Ok, I'll mark as `= delete`, when I get back to a computer.

> Have you tested on a machine with a battery? I haven't, but I can.

Sorta. It was plugged in, and might be more help if I got a part of a trace where the battery is discharging.

...that reminds me, the temperatures that were recorded in the trace seemed out of whack, I should investigate that.

Also: Is there any documented limit to the number of batteries? I hate using an arbitrary number (`5`).


> I'm not convinced that setting up deviceIFaceDetailDataByteSize in order to avoid one /analyze warning is a good tradeoff. /analyze should make code better, not messier. I would have just used #pragma warning(suppress). But, I don't care that much either way.

Eh? I don't think *that* was to suppress a warning? That was to factor that part of the code out, *partly* so that I could stick another `const` in that big loop.

(*more to come*)

randomascii:
I don't think there is a real limit on the number of batteries, although practically five is unlikely to be exceeded. I think the maximum is to prevent looping forever if the API is buggy - too many drivers are.

The temperatures sometimes show up in Fahrenheit. It's quite weird and I'm not sure what the pattern is.

I guess the purpose of deviceIFaceDetailDataByteSize wasn't completely clear. I get nervous anytime that many lines of code are added for no obvious purpose. Adding const is nice and can avoid bugs, but increasing lines-of-code increases the odds of bugs hiding (all else being equal) so I'm not sure that extracting that function is worthwhile?

ariccio:
> There are some style inconsistencies (parentheses on the same line on struct and namespace declarations) but I'm not sure how much I care about them.

Eh?


> I guess the purpose of deviceIFaceDetailDataByteSize wasn't completely clear. I get nervous anytime that many lines of code are added for no obvious purpose. Adding const is nice and can avoid bugs, but increasing lines-of-code increases the odds of bugs hiding (all else being equal) so I'm not sure that extracting that function is worthwhile?

Well, that function [wasn't *strictly new* code](https://github.com/google/UIforETW/blob/master/UIforETW/PowerStatus.cpp#L119), all I really did was  add the sanity check for the unexpected success of `SetupDiGetDeviceInterfaceDetail`, and it's only ~10 lines at the moment. If you'd really like, I can merge it back in to the loop.

randomascii:
UIforETW puts parentheses on the next line whereas the new structs and namespace you add have the open paren at the end of the line.

I haven't really checked to see how much complexity it really adds to have SetupDiGetDeviceInterfaceDetail out-of-line, it just feels more complex - I guess I like straightline code. But, I'm fine either way.

ariccio:
> There are some style inconsistencies (parentheses on the same line on struct and namespace declarations) but I'm not sure how much I care about them.

Eh?


> I guess the purpose of deviceIFaceDetailDataByteSize wasn't completely clear. I get nervous anytime that many lines of code are added for no obvious purpose. Adding const is nice and can avoid bugs, but increasing lines-of-code increases the odds of bugs hiding (all else being equal) so I'm not sure that extracting that function is worthwhile?

Well, that function [wasn't *strictly new* code](https://github.com/google/UIforETW/blob/master/UIforETW/PowerStatus.cpp#L119), all I really did was  add the sanity check for the unexpected success of `SetupDiGetDeviceInterfaceDetail`, and it's only ~10 lines at the moment. If you'd really like, I can merge it back in to the loop.

ariccio:
Grr. This DSL is *that* slow - I intended to post before that last comment of yours.

ariccio:
> UIforETW puts parentheses on the next line whereas the new structs and namespace you add have the open paren at the end of the line.

Uhhh... braces?

ariccio:
Update: `deviceIFaceDetailDataByteSize` has a code complexity of two:
deviceifacedetaildatabytesize

randomascii:
Yes - braces.

ariccio:
I'm sorry, my schedule has changed - I may not be at my computer for nearly a week - should I just wait for a week, or maybe you could squash and tweak yourself?

randomascii:
I can do that. I'll be sure to mention that most of the changes came from you.

ariccio:
Ok. Fire at will.

randomascii:
I looked into deviceIFaceDetailDataByteSize and decided to make it not a function, for two reasons.

1) It forces the reader to move around more in order to understand the code. It's a judgment call and if that was the only reason I'd leave it.
2) The function has a very leaky abstraction. The caller has to check GetLastError(). That also means that deviceIFaceDetailDataByteSize actually returns bytesNeeded and *then* checks whether it is valid. That is, the function potentially reads uninitialized data. Practically speaking I doubt that it does, but contractually that is what can happen.

I also suppressed the remaining /analyze warning.

PR #44, opened 2015-07-30
etwprof.h: Adjust guard to only enable under MSVC, and allow a disable define

rpavlik:
The previous code conflates building for Windows with building with MSVC, while we'd like consuming code to be able to build with other compilers on Windows too.

### DIFF
--- a/include/ETWProviders/etwprof.h
+++ b/include/ETWProviders/etwprof.h
@@ -30,7 +30,7 @@ limitations under the License.
 
 typedef long long int64;
 
-#ifdef	_WIN32
+#if defined(_WIN32) && defined(_MSC_VER) && !defined(ETW_DISABLE)
 // ETW support should be compiled in for all Windows PC platforms. It isn't
 // supported on Windows XP but that is determined at run-time. This #define
 // is used to let the code compile (but do nothing) on other operating systems.


randomascii:
Why do you only want ETW enabled under MSVC? ETW is available under Windows and the compiler used should be irrelevant. Adding a #define to allow explicit disabling could be helpful.

Note that before accepting pull requests I need you to sign the CLA -- see the instructions at https://github.com/google/UIforETW/blob/master/CONTRIBUTING

rpavlik:
Err... so I guess that I:
1. glossed over the fact that compiler != system sdk given that now I can use clang in visual studio
2. assumed that mingw didn't have the required headers, probably because I...
3. saw the SAL annotations and figured those wouldn't work without MSVC (I've typically done project-specific wrapping of SAL annotations I want to use that define them typically to nothing on non-msvc, though I suppose I haven't actually tested building with them left in on a modern mingw or clang-cl

Oops.

I think the only issue there might be the SAL annotations, because if nothing else it's a C API so I should be able to use an MSVC-built library with mingw even if the headers used by the implementation are missing there.

Thought I signed a CLA back in the day (for IWYU) but I see new web app since then, and new employer, so will go through that again.

randomascii:
I looked up your github username and email address and saw no sign of a CLA. Sorry about the requirement, but it is a requirement.

SAL annotations are easy to support with other compilers because they can be preprocessed to nothing - at least I think that is how it works. Either that or they can be ignored by the compiler. That's not to say that they *are* supported, but I wouldn't be surprised if clang-on-Windows can handle having them there.

rpavlik:
OK, there should now be a CLA through my employer (firstname at sensics.com).

I guess I'll have to test the SAL thing some more.

ariccio:
> SAL annotations are easy to support with other compilers because they can be preprocessed to nothing

Yup. (the way that it's usually done is to) Create a file called `NoSAL.h` (or whatever), and just `#define` all the macros as nothing.

I assume thang Clang-on-windows does something similar, else it'd have some trouble with `Windows.h`.

rpavlik:
OK, so with a dummy SAL (<https://github.com/OSVR/OSVR-Core/tree/master/vendor/dummy-sal>), and now that it's all `extern "C"` I can build against it (though not build it - can't compile the instrumentation manifest) with MinGW, closing this out.

PR #46, opened 2015-08-07
Ignoring auto-generated Incremental LTCG files, updated AUTHORS.

mwinterb:
Incremental Link Time Code Generation was introduced in VS2015 to allow for only partial recompilation during the link phase. The incremental status files are stored alongside the target exe / dll, and there doesn't seem to be a trivial way to force them into the Intermediate Directory directory.


mwinterb:
See [this VCBlog post](http://blogs.msdn.com/b/vcblog/archive/2014/11/12/speeding-up-the-incremental-developer-scenario-with-visual-studio-2015.aspx) for the overview of Incremental LTCG, if you hadn't already. I would have assumed that James McNellis's StackOverflow answer [here](http://stackoverflow.com/a/31566092) would have mentioned if there was a specific way of changing the output location, but this may just be like the normal incremental linking ilk files without a way of customizing it. The [current docs](https://msdn.microsoft.com/en-us/library/xbf3tbeh.aspx) don't even mention the option.

I could have put this in the root .gitignore file, but my personal preference is to make ignore files as precise as possible, and since there was already one in the only directory that could have the new files, this way seemed best. Even though the *.ilk ignore is in the root one, and these are analogous to those files. Um. I can just make that change if you'd like and update this.

randomascii:
As more projects get added we may want to move the .ignore filters to the root level, but they are fine in bin for now.

PR #47, opened 2015-08-08
Free EnergyLib.dll when CPowerStatusMonitor is being destroyed.

mwinterb:


mwinterb:
I'm a bit wishy-washy on this change just because even if someone wanted to extract CPowerStatusMonitor for use in a non-UIforETW environment and would want to free the resources allocated by EnergyLib prior to process exit, it seems like EnergyLib isn't really designed for completely independent components using it simultaneously. For instance, calling IntelEnergyLibInitialize twice results in a heap corruption in the first call to ReadSample. At least the samples appear to be thread-local so the separate components could protect themselves that way once everything's been initialized.

I realize this concern is a bit academic given its current implementation (since two components each generating ETW events for the same data set would be a bit odd), I was hoping to separate the data gathering from the data pushing.

randomascii:
Yeah, mostly academic, but not unreasonable.

I realized after I pulled the change that maybe I should have suggested omitting the != nullptr part of it, but hey, whatevs.

mwinterb:
I mainly put the conditional in since FreeLibrary is annotated with _In_ and not _In_opt_.

randomascii:
What I meant was that "if (energyLib_) FreeLibrary(energyLib_);" would be preferable - not omitting the test, but omitting the nullptr part of the test.

Personal taste, not a big deal either way.

mwinterb:
Oh, right. Force of habit, but that should have been obvious given the Init change. I'll keep that in mind.

PR #50, opened 2015-08-19
Add extern "C" to the plain C functions

rpavlik:
With a dummy `sal.h` header, I can get a client of `etwproviders.dll` to compile in mingw. However, the "plain" functions weren't marked `extern "C"` so linking fails due to mismatched name mangling.

This patch adds `extern "C"` to those functions. As long as I was adding `#ifdef __cplusplus` I wrapped the class (both the real and fake definitions) in it as well.

Should have a CLA from my company now.

### DIFF
--- a/include/ETWProviders/etwprof.h
+++ b/include/ETWProviders/etwprof.h
@@ -50,6 +50,10 @@ const int kFlagDoubleClick = 100;
 #endif
 #include <sal.h> // For _Printf_format_string_
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 // Insert a single event to mark a point in an ETW trace.
 PLATFORM_INTERFACE void __cdecl ETWMark(_In_z_ PCSTR pMessage);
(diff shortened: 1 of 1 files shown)


randomascii:
Interesting timing. Just today I was thinking of adding an ETWMark overload, which would be incompatible with this idea. I can always call it ETWMarkW or some-such (it will use a wchar_t* parameter).

One disadvantage of this change is that it is a breaking change - all clients will need to be rebuilt. I'm not sure how serious a problem that is. Let me ponder it.

I confirmed the CLA.

randomascii:
Merged. You should do another pull request to add yourself to the CONTRIBUTORS and AUTHORS file. Thanks for contributing!

ariccio:
@rpavlik I see that you've put this [to good use](https://osvr.github.io/presentations/20150901-Intro-ETW-OSVR.pdf)!

rpavlik:
Yes, nice to be able to build that with msvc and mingw, and nice to be able to publicly share use of such a good tool 😀

PR #56, opened 2015-10-07
cleanup some utility functions

ariccio:
I cleaned up the utility functions while tracking down some bad behavior of WPA profile copying.

There are minimal functional changes, where I replaced unsupported functions (`SHGetSpecialFolderPath`) with recommended functions (`SHGetKnownFolderPath`), switched custom functions (`GetWindowsVersion`) to their built-in equivalents (`IsWindows7OrGreater`, etc...), and set up some error handling.

As you prefer no-fail functions, I've wrapped several Win32 APIs so that when failure is reasonable, they simply log (via `debugPrintf`) it, or *else* (failure *is* a logic bug) they `std::terminate()`. For more complex APIs, I've separated the error logging code itself from the caller, so that the caller only checks if the wrapper returned  `true`/`false`.

Other changes:
- I wrote the very simple `_wgetenv_s`/`getenv_s` wrapper that returns a `std::wstring`/`std::string`, so you don't need to `#pragma warning(suppress : 4996)` anymore
- I rewrote `GetFileExt` to use the `std::wstring` algorithms (`find_last_of`, etc...) instead of the `C` algorithms. The rewritten `C++`-style functions are easier to read (I think so).

The result of the cleanup is 403 additional lines of code, but I think it's easier to read, and more reliable. It's your choice to accept this, but it fixes the odd behavior that initiated this, and should make future development easier.

I'm not exactly sure why git squashed my eight commits down to two commits. If it's a problem, I'll happily redo it.


UIforETW/Support.cpp
```diff
@@ -167,8 +167,8 @@ int CUIforETWDlg::BufferCountBoost(int requestCount) const
 	if (!GlobalMemoryStatusEx(&memoryStatus))
 		return requestCount;
 
-	const int64_t oneGB = int64_t(1024) * 1024 * 1024;
-	int64_t physicalRam = memoryStatus.ullTotalPhys;
+	const int64_t oneGB = 1024i64 * 1024i64 * 1024i64;
```
randomascii: 1024i64 is almost unreadable I think. At a glance it reads like 1024164 which is a very strange constant.

UIforETW/UIforETWDlg.cpp
```diff
@@ -282,7 +280,7 @@ BOOL CUIforETWDlg::OnInitDialog()
 		}
 	}
 
-	if (GetWindowsVersion() == kWindowsVersionXP)
+	if (!IsWindowsVistaOrGreater())
```
randomascii: Is there a WindowsXPOrLesser() function? Ditto for Windows7OrLesser()? The negatives in all of these expressions make the code harder to read.

UIforETW/Support.cpp
```diff
@@ -167,8 +167,8 @@ int CUIforETWDlg::BufferCountBoost(int requestCount) const
 	if (!GlobalMemoryStatusEx(&memoryStatus))
 		return requestCount;
 
-	const int64_t oneGB = int64_t(1024) * 1024 * 1024;
-	int64_t physicalRam = memoryStatus.ullTotalPhys;
+	const int64_t oneGB = 1024i64 * 1024i64 * 1024i64;
+	const int64_t physicalRam = memoryStatus.ullTotalPhys;
```
randomascii: Yay const!

UIforETW/UIforETWDlg.cpp
```diff
@@ -1599,7 +1590,7 @@ void CUIforETWDlg::OnContextMenu(CWnd* pWnd, CPoint point)
 				pContextMenu->EnableMenuItem(id, MF_BYCOMMAND | MF_GRAYED);
 		}
 
-		if (GetWindowsVersion() < kWindowsVersion8)
+		if (IsWindows8OrGreater())
```
randomascii: The sense of this is backwards - i.e.; this is a bug.

UIforETW/Utility.cpp
```diff
@@ -19,7 +19,166 @@ limitations under the License.
...
+PCWSTR const WPAStartupFileName = L"\\Startup.wpaProfile";
+
+std::wstring GetDocumentsFolderPath()
+{
+	//must CoTaskMemFree when done!
+	PWSTR docsPathTemp = NULL;
+	const HRESULT docsPathResult = ::SHGetKnownFolderPath(FOLDERID_Documents, KF_FLAG_NO_ALIAS, NULL, &docsPathTemp);
+	if (FAILED(docsPathResult))
+	{
+		if (docsPathResult == E_INVALIDARG)
```
randomascii: How does the error handling here add value? If we can't get the documents folder path then can we really continue? Why not delete 15 lines of code and just terminate if the docs folder can't be found? If the error handling adds value I want to know what it is.

UIforETW/Utility.cpp
```diff
@@ -19,7 +19,166 @@ limitations under the License.
...
+
+void copyWPAProfileToDocuments(const bool force)
+{
+	// First copy the WPA 8.1 startup.wpaProfile file
+	std::wstring docsPath(GetDocumentsFolderPath());
+
+	if (force)
+		outputPrintf(L"\n");
+
+	if (docsPath.empty())
```
randomascii: Why are we handling this?

UIforETW/Utility.cpp
```diff
@@ -594,69 +871,171 @@ void SetCurrentThreadName(PCSTR threadName)
...
-				if (force)
-					outputPrintf(L"%s", L"Copied Startup.10wpaProfile to %localappdata%\\Windows Performance Analyzer\n");
-			}
-			else
-			{
-				if (force)
-					outputPrintf(L"%s", L"Failed to copy Startup.10wpaProfile to %localappdata%\\Windows Performance Analyzer\n");
-			}
-		}
+bool ClipboardEmpty()
```
randomascii: This is a badly named function. "ClipboardEmpty()" sounds like it queries the clipboard to see if it is empty. It does not sound like it empties the clipboard.

UIforETW/Utility.cpp
```diff
@@ -594,69 +871,171 @@ void SetCurrentThreadName(PCSTR threadName)
...
+void ClipboardClose()
+{
+	const BOOL closeResult = ::CloseClipboard( );
+	if (closeResult)
+		return;
+	debugPrintf(L"Failed to close the clipboard!\n");
+	debugLastError( );
+}
+
+bool ClipboardOpen()
```
randomascii: This name does not suggest that it opens the clipboard. It sounds like it queries the open status of the clipboard.

UIforETW/Utility.cpp
```diff
@@ -594,69 +871,171 @@ void SetCurrentThreadName(PCSTR threadName)
...
 	}
+	(*hmem) = mem;
+	return true;
+}
 
+void FreeGlobalMemory(_Frees_ptr_opt_ const HGLOBAL hmem)
+{
+		//If [GlobalFree] succeeds, the return value is NULL.
+		const HGLOBAL freeRes = ::GlobalFree(hmem);
+		if (freeRes != NULL)
```
randomascii: Every line of code has a cost. It's not huge, but it's there. Is the odds of failure of GlobalFree high enough to justify checking for this case and terminating? If GlobalFree does fail does calling std::terminate() help the situation in any way?

UIforETW/Utility.cpp
```diff
@@ -351,75 +603,112 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
+	const int result = ::SHFileOperationW(&fileOp);
 
 	return result;
 }
 
 void SetClipboardText(const std::wstring& text)
 {
-	BOOL cb = OpenClipboard(GetDesktopWindow());
-	if (!cb)
+	if (!ClipboardOpen())
```
randomascii: What is the advantage here? It saves a line of code but it makes the code harder to read (what does ClipboardOpen() do?) and spreads out the logic.

UIforETW/Utility.h
```diff
@@ -134,4 +140,31 @@ void SetCurrentThreadName(PCSTR threadName);
...
+bool ClipboardEmpty();
+
+_Success_(return)
+bool ClipboardSetUnicode(_In_ const HGLOBAL hmem);
+
+_Success_(return)
+bool ClipboardGetUnicode(_Out_ _Post_valid_ HANDLE* const hClip);
+
+_Success_(return)
+bool AllocGlobalMemoryForString(const std::wstring& text, _Out_ _Post_valid_ HGLOBAL* const hmem);
```
randomascii: Does this function (and many other newly added functions) deserve to be in utility.h? Is it generally useful outside of utility.cpp? I think not and it makes the high-level useful functions harder to find.

UIforETW/Utility.h
```diff
@@ -108,24 +114,24 @@ class ElapsedTimer
...
-		QueryPerformanceCounter(&start_);
+		ATLVERIFY(QueryPerformanceCounter(&start_));
 	}
 	double ElapsedSeconds() const
 	{
-		LARGE_INTEGER stop;
-		QueryPerformanceCounter(&stop);
-		LARGE_INTEGER frequency;
-		QueryPerformanceFrequency(&frequency);
+		LARGE_INTEGER stop = {0};
```
randomascii: Should be  "= {};" rather than "= {0};"

UIforETW/Utility.h
```diff
@@ -76,30 +88,24 @@ int64_t GetFileSize(const std::wstring& path);
...
-{
-	kWindowsVersionXP,
-	kWindowsVersionVista,
-	kWindowsVersion7,
-	kWindowsVersion8,
-	kWindowsVersion8_1,
-	kWindowsVersion10,
-};
+std::wstring GetEnvironmentVariableString(_In_z_ PCWSTR variable);
+std::string GetEnvironmentVariableString(_In_z_ PCSTR variable);
```
randomascii: I don't think a narrow-character version of this is actually needed. The narrow-char usages of getenv are purely accidental.

randomascii:
There's some good stuff there, but also some excessive error checking I think. We're not trying to unit-test Windows, and we're not trying to continue running even on a Windows install that doesn't have a documents directory.

SetClipboardText, for instance, gets significantly longer and it relies on a half-dozen new functions. This makes reading the code much harder (what does ClipboardOpen() do?) and it is not clear to me what the value is. There has to be a positive value-per-line-of-code ratio.

CopyStartupProfiles is odd. Breaking it up is okay, I guess. The extra error checking seems excessive, and I'm not sure why the two helper functions are more than 700 lines from the one place that they are used. Locality of reference is handy for readers.

I would recommend smaller more tightly focused pull requests. The large-scale rearrangements don't clearly add value, whereas adding const, ATLVERIFY, final, and GetEnvironmentVariableString() are clearly good.

Anyway, I hope the feedback is helpful. I'd love to get some of the changes incorporated.

UIforETW/UIforETWDlg.cpp
ariccio: [Your comment below](https://github.com/google/UIforETW/pull/56/files#r41416669) is a case in point.

UIforETW/UIforETWDlg.cpp
ariccio: Fixed.

UIforETW/Support.cpp
ariccio: Reverted. I got carried away with my integer literals :)

UIforETW/Utility.cpp
ariccio: You know, I was thinking that some computers might *not* have a Documents folder... but in retrospect, that's just plain silly. This is just pathological thoroughness. Removed.

UIforETW/Utility.cpp
ariccio: It's actually no longer relevant now that I've removed the error path from `GetDocumentsFolderPath`.

UIforETW/Utility.cpp
ariccio: Hmm. That *IS* terrible. What the hell was wrong with me?

randomascii:
I'd definitely like some of the changes with clear benefits - maybe separate them out into a separate pull request, or separate the more controversial (added error checking) changes into a separate pull request? I'd definitely like the const, ATLVERIFY, final, and GetEnvironmentVariableString() changes.

UIforETW/Utility.h
randomascii: Actually, the initializers and verifications should just go away. The QueryPerformanceFrequency documentation on MSDN says:
"On systems that run Windows XP or later, the function will always succeed and will thus never return zero."

No point checking for errors that can never happen.

ariccio: Huh. I never noticed that. They should probably stick a couple of `_Always_` annotations on it.

For reference: [QueryPerformanceFrequency function](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644905.aspx)

UIforETW/Utility.h
ariccio: I'd need to change `CUIforETWDlg::SetSymbolPath` to use wide character versions. Should I?

randomascii: Ah - I somehow missed that one when I was looking to see if a narrow character version was ever needed.

If there aren't any added complexities in doing that then that would be the right way to go. That would also get rid of a call to AnsiToUnicode().

However there may be a long tail of other changes needed - systemDrive_, for instance, is a string. However I think that's basically it for std::string, and almost it for AnsiToUnicode.

UIforETW/Utility.h
ariccio: I'm actually not sure why I didn't put these in an anonymous namespace (in the `.cpp` file). Fixed.

`outputLastError` and `debugLastError` *might* be useful elsewhere, but I can totally remove them from the header if you'd like.

UIforETW/UIforETWDlg.cpp
ariccio: I just wrote them. Fixed.

UIforETW/Utility.cpp
ariccio: My intent was to make it look obviously different from the Windows `::OpenClipboard()` function, but that didn't work so well. I've reversed it (and the other Clipboard... functions).

UIforETW/Utility.cpp
ariccio: I believe that if `::GlobalFree` fails, it's the result of a serious logic bug? I could instead call `outputLastError` to make the error visible to the user without killing the whole process.

UIforETW/Utility.cpp
```diff
@@ -126,71 +459,126 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
+
+	// Double-verify that the buffer is null-terminated.
+	buffer[buffer.size() - 1] = 0;
+	std::wstring result = &buffer[0];
 	return result;
 }
 
 // Return a string from a format string and some printf-style arguments.
-std::wstring stringPrintf(_Printf_format_string_ const wchar_t* pFormat, ...)
+std::wstring stringPrintf(_Printf_format_string_ PCWSTR const pFormat, ...)
```
ariccio: We actually [discussed this in a previous PR (the "ETW providers cleanup")](https://github.com/google/UIforETW/pull/33#discussion_r30299346), and I tried to set up a test for this the other day, but `/analyze` was insanely flaky. It refused to even pick up obvious non-null terminated strings. If you're curious, or just need a reason to be angry at /analyze, you can check it out here: https://github.com/ariccio/test-printf-format-string-annotation/

UIforETW/Utility.h
ariccio: I think I'll leave this to another PR? This is a tiny bit big already.

ariccio:
I've addressed most of your comments, but now the PR is messed up. Grr. I'll redo it.

PR #60, opened 2015-10-15
Removed usage of MFC "feature pack" classes.

mwinterb:
The primary goal of this is to remove dependencies on DLLs that are not included in the list of DLLs that are supported on the Windows Server Core configurations. This removes the dependencies on MSIMG32.dll, gdiplus.dll, and WINMM.dll. It also gets rid of a dependency on imm32.dll, but that one is actually supported.

As a side effect, this also reduces the static exe size by about 2 megabytes, but the actual benefit of that ends up being minimal.

### DIFF
--- a/UIforETW/Settings.cpp
+++ b/UIforETW/Settings.cpp
@@ -18,7 +18,6 @@ limitations under the License.
 #include "UIforETW.h"
 #include "Settings.h"
 #include "Utility.h"
-#include "afxdialogex.h"
 
 /*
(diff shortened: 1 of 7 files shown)


randomascii:
Hey, that's great! 2 MB smaller (per binary, so 4 MB total) is great, and four fewer DLLs being imported. Much appreciated.

k-y:
Awesome work guys, thanks for the effort on Server Core configs

PR #62, opened 2015-10-16
Tweak utility funcs try four, redo of #56

ariccio:
Phew, that took a couple tries. I've addressed nearly everything discussed in #56, and got git's history straightened out. I could probably squash it down to a single commit, but it was enough trouble to squash it down to two commits ((one commit of my changes) + (one pull from & resolve conflicts with master)).


UIforETW/Utility.cpp
```diff
@@ -20,7 +20,305 @@ limitations under the License.
...
+	{
+		const BOOL makeDirResult = ::CreateDirectoryW(destDir.c_str(), NULL);
+		if (makeDirResult == 0)
+		{
+			const DWORD lastErr = ::GetLastError();
+
+			//ERROR_ALREADY_EXISTS is reasonable.
+			if (lastErr != ERROR_ALREADY_EXISTS)
+			{
+				outputPrintf(L"Something went wrong when copying the WPA profile to the documents folder.\n");
```
randomascii: I still think you're going overboard with error checking. This change adds 384 lines to utility.cpp - making it 57% longer - with much of that being checking for errors that are either impossible or unimportant. What is the value in telling the user "Something went wrong" as opposed to just letting CopyFileW fail, for example.

UIforETW/Utility.cpp
```diff
@@ -20,7 +20,305 @@ limitations under the License.
...
+				outputLastError(lastErr);
+				return;
+			}
+		}
+		const BOOL copyResult = ::CopyFileW(source.c_str(), dest.c_str(), FALSE);
+		if (copyResult)
+		{
+			//why does the message depend on forced copy?
```
randomascii: If the startup profile happened automatically at UIforETW startup then the user would only be confused about error messages.

UIforETW/Utility.cpp
```diff
@@ -20,7 +20,305 @@ limitations under the License.
...
+	debugLastError();
+	return false;
+}
+
+void CloseRegKey(_In_ _Pre_valid_ _Post_ptr_invalid_ const HKEY key, PCWSTR const keyName)
+{
+	const LONG closeKey = ::RegCloseKey(key);
+	if (closeKey == ERROR_SUCCESS)
+		return;
+	outputPrintf(L"Failed to close registry key `%s`.\n", keyName);
```
randomascii: What is the value added by logging this? I would argue that it is negative value. If a user sees this then they will be confused, and any user-targeted messages will be obscured by this one.

UIforETW/Utility.cpp
```diff
@@ -20,7 +20,305 @@ limitations under the License.
...
+	}
+	return true;
+}
+
+bool emptyClipboard()
+{
+	const BOOL emptyClip = ::EmptyClipboard();
+	if (!emptyClip)
+	{
+		outputPrintf(L"Failed to empty clipboard!\n");
```
randomascii: Value of this error checking is...?

UIforETW/Utility.cpp
```diff
@@ -127,71 +460,126 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
-	if (result == ERROR_SUCCESS)
+	if (!OpenRegKey(&key, root, subkey.c_str()))
+		return;
+
+	const LONG setResult = ::RegSetValueExW(key, valueName.c_str(), 0, REG_DWORD, reinterpret_cast<const BYTE*>(&value), sizeof(value));
+	if (setResult != ERROR_SUCCESS)
 	{
-		RegSetValueEx(key, valueName.c_str(), 0, REG_DWORD, reinterpret_cast<const BYTE*>(&value), sizeof(value));
-		RegCloseKey(key);
+		outputPrintf(L"Failed to set registry key `%s`.\n", subkey.c_str());
```
randomascii: Is it actually likely that this RegSetValueExW key will fail on customer machines? If logic errors are possible then an assert could make sense, but outputPrintf is communicating with the user.

UIforETW/Utility.cpp
```diff
@@ -278,59 +664,64 @@ void SmartEnableWindow(HWND Win, BOOL Enable)
...
+		return path.substr(0, lastSlash+1);
+	}
 	// If there's no slash then there is no directory.
 	return L"";
 }
 
 std::wstring CrackFilePart(const std::wstring& path)
 {
-	std::wstring filePart = GetFilePart(path);
+	const std::wstring filePart = GetFilePart(path);
```
randomascii: const is always welcome.

UIforETW/Utility.cpp
```diff
@@ -352,75 +743,112 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
 	{
-		void *ptr = GlobalLock(hmem);
-		if (ptr != NULL)
-		{
-			memcpy(ptr, text.c_str(), length);
-			GlobalUnlock(hmem);
+		closeClipboard();
+		return;
+	}
+	const HANDLE& hmem = hmem_temp;
```
randomascii: Why do we need a second name for hmem_temp? I don't understand.

UIforETW/Utility.cpp
```diff
@@ -352,75 +743,112 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
+
+	size_t bytes_temp = 0u;
+	if (!SizeGlobalMemory(hClip, &bytes_temp))
+	{
+		UnlockGlobalMemory(hClip);
+		closeClipboard();
+		return result;
+	}
+
+	const size_t bytes = bytes_temp;
```
randomascii: Why copy the value? const variables are nice, but only having one copy of a value is also nice.

UIforETW/Utility.cpp
```diff
@@ -352,75 +743,112 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
+		debugPrintf(L"Failed to get file size!\n");
+		debugLastError();
 		return 0;
+	}
 
-	if (GetFileSizeEx(hFile, &result))
-	{
-		CloseHandle(hFile);
+	const BOOL gfsResult = ::GetFileSizeEx(hFile, &result);
+	CloseValidHandle(hFile);
```
randomascii: Probably off topic, but instead of CloseValidHandle wouldn't it be better to have a ScopedHandle class that would close the handle automatically (and do any appropriate error checking)?

UIforETW/Utility.cpp
```diff
@@ -431,108 +859,116 @@ bool Is64BitWindows()
 #else
 	// http://blogs.msdn.com/b/oldnewthing/archive/2005/02/01/364563.aspx
 	BOOL f64 = FALSE;
-	bool bIsWin64 = IsWow64Process(GetCurrentProcess(), &f64) && f64;
+	const bool bIsWin64 = ( ::IsWow64Process(::GetCurrentProcess(), &f64) && f64 );
```
randomascii: The outer set of parentheses seem odd - there's no precedence for them to clarify.

UIforETW/Utility.cpp
```diff
@@ -431,108 +859,116 @@ bool Is64BitWindows()
...
 }
 
 bool Is64BitBuild()
 {
+	//when MSVC gets constexpr, this is a perfect candidate.
 	return sizeof(void*) == 8;
 }
 
-WindowsVersion GetWindowsVersion()
+bool IsWindowsTenOrGreater()
```
randomascii: Nice.

UIforETW/Utility.h
```diff
@@ -39,11 +39,23 @@ void CreateRegistryKey(HKEY root, const std::wstring& subkey, const std::wstring
...
 
+
+//MultiByteToWideChar: https://msdn.microsoft.com/en-us/library/windows/desktop/dd319072.aspx
+//
+//Remarks:
+//As mentioned in the caution above,
+//the output buffer can easily be overrun
+//if this function is not first called with cchWideChar set to 0
+//in order to obtain the required size. 
+int RequiredNumberOfWideChars(const std::string& text);
```
randomascii: This shouldn't be exported in utility.h - it is an internal detail of utility.cpp.

UIforETW/Utility.h
```diff
@@ -121,11 +130,10 @@ class QPCElapsedTimer
 		QueryPerformanceCounter(&stop);
 		LARGE_INTEGER frequency;
 		QueryPerformanceFrequency(&frequency);
-
 		return (stop.QuadPart - start_.QuadPart) / float(frequency.QuadPart);
 	}
 private:
-	LARGE_INTEGER start_;
+	LARGE_INTEGER start_ = {};
```
randomascii: The = {} isn't needed and will just waste code space.

UIforETW/Utility.cpp
```diff
@@ -20,7 +20,305 @@ limitations under the License.
...
+	outputPrintf(L"Failed to close registry key `%s`.\n", keyName);
+	debugLastError();
+}
+
+void CloseFindHandle(_In_ _Pre_valid_ _Post_ptr_invalid_ const HANDLE handle, PCWSTR const directory)
+{
+	const BOOL findClose = ::FindClose(handle);
+	if (findClose)
+		return;
+	outputPrintf(L"FindClose (for directory: `%s`) failed.\n", directory);
```
randomascii: I think this wrapping adds very little value. Either wrap it in a class, or don't wrap it. See ScopedHandle in Chromium.

UIforETW/Utility.cpp
```diff
@@ -20,7 +20,305 @@ limitations under the License.
...
+PCWSTR const WPAStartupFileName = L"\\Startup.wpaProfile";
+
+_Success_(return)
+bool AllocGlobalMemoryForString(const std::wstring& text, _Out_ _Post_valid_ HGLOBAL* const hmem)
+{
+	const size_t length = (text.size() + 1) * sizeof(text[0]);
+	const HANDLE mem = ::GlobalAlloc(GMEM_MOVEABLE, length);
+	if (!mem)
+	{
+		outputPrintf(L"Couldn't allocate memory to set clipboard text!\n");
```
randomascii: This is supposed to be a generic allocation function so its error messages shouldn't reference a specific use-case.

UIforETW/Utility.cpp
```diff
@@ -352,75 +743,112 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
 	};
 	// Delete using the recycle bin.
-	int result = SHFileOperation(&fileOp);
+	//TODO: IFileOperation?
+	const int result = ::SHFileOperationW(&fileOp);
 
 	return result;
 }
 
 void SetClipboardText(const std::wstring& text)
```
randomascii: I think that SetClipboardText and GetClipboardText and their associated helper functions are my least favorite changes in this pull request. They add a *ton* of code, they add an extra level of indirection to everything, they make the code harder to read (extra level of indirection) and it's not clear what value they add. The original functions shouldn't ever fail, but they had error checking just in case.

I think you understand the cognitive cost of increased lines of code, increased indirection (function calls instead of straightline code) and custom 'helper' functions that are non-standard and must be learned by anybody who wants to review the code. The original code was simple, handled most errors (could handle a couple more), and could be understood by any Win32 programmer with little effort. The new version is far more complicated and I still don't know what the added value is.

randomascii:
It's weird that it won't squash to one commit - some days I just don't understand git - but that doesn't really matter. Various comments left.

UIforETW/Utility.cpp
```diff
@@ -127,71 +460,126 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
 
-std::wstring GetEditControlText(HWND hEdit)
+std::wstring GetEditControlText(const HWND hEdit)
 {
-	std::wstring result;
-	int length = GetWindowTextLength(hEdit);
+	const int length = ::GetWindowTextLengthW(hEdit);
 	std::vector<wchar_t> buffer(length + 1);
-	GetWindowText(hEdit, &buffer[0], static_cast<int>(buffer.size()));
+
```
randomascii: Trailing tab here...

UIforETW/Utility.cpp
```diff
@@ -127,71 +460,126 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
+	//GetWindowText https://msdn.microsoft.com/en-us/library/windows/desktop/ms633520.aspx
+	//If [GetWindowTextW] succeeds, the return value is the length,
+	//in characters, of the copied string, not including the
+	//terminating null character.
+
+	//If the window has no title bar or text,
+	//[or] if the title bar is empty,
+	//or if the window or control handle is invalid,
+	//the return value is zero. 
+
```
randomascii: Trailing tab here also. Only really visible in "git show a6dcff1"

randomascii:
This talk on the importance of simplicity seems relevant:

http://www.infoq.com/presentations/Simple-Made-Easy

"Simplicity is a prerequisite for reliability" - Dijkstra
"Your ability to reason about your program is critical to debugging"

Interesting thoughts on what matters.

randomascii:
BTW, if you don't want to work on this anymore I can merge it in and then revert the parts that I'm not happy with, which would leave most of it intact and all of it in the history.

ariccio:
I think I'll get to it tonight.

sent from my (stupid) windows phone

-----Original Message-----

UIforETW/Utility.cpp
ariccio: Yup, I was "programming on quicksand". It's better to just let it fail. I'm removing. (should I log a debug message?)

UIforETW/Utility.cpp
ariccio: I can't actually remember why I decided to *log* this instead of `std::terminate()`ing, or just ignoring it. A failed `RegCloseKey` either means:

1. A logic error (like `CloseHandle`)
2. We're leaking a handle, and may do so several times:
uiforetw_regclosekey

...it may be totally ok to leak a few registry key handles, but I prefer not to. Should I just remove the check?

ariccio: ...or, alternatively, `UIETWASSERT` it?

UIforETW/Utility.cpp
ariccio: Logging is probably unnecessary (like above), but a failure to empty the clipboard will mean either:

1. `SetClipboardData` fails later, for mysterious reasons
2. The clipboard may - and this is a guess - end up with garbage data of some sort.

...should I also `std::terminate()` or `UIETWASSERT` this? (thus the function would become a `void` function).

UIforETW/Utility.cpp
ariccio: Here, I think the real issue is that if `RegSetValueEx` fails, we might not be able to set `DisablePagingExecutive`, and then xperf/ETW stack walking wouldn't work. Since ETW is vastly less useful without stack walking, UIforETW is then also much less useful. 

In that case, we'd probably be better off handling that in `CUIforETWDlg::DisablePagingExecutive` instead. Otherwise I can `UIETWASSERT` this instead.

UIforETW/Utility.cpp
ariccio: Removed.

UIforETW/Utility.cpp
ariccio: Removed.

UIforETW/Utility.cpp
ariccio: Not really off topic: yup. Should I add it in this PR or another?

UIforETW/Utility.cpp
ariccio: A couple of reasons, which are probably moot:

1. (first and foremost is that) I consciously treat any multi-operator expression as suspicious. If it's not totally obvious, I'll always look up [C++ operator precedence](http://en.cppreference.com/w/cpp/language/operator_precedence), to be sure. Adding parenthesis makes precedence explicit, and eliminates any concern for the [insidiousness of C](http://www.underhanded-c.org/_page_id_2.html).
2. The historical basis for the first reason is in my copy of MISRA C (section 8, rule 12.1), which gives as an Advisory guideline ("should be followed as far as is reasonably practical") that expressions containing any of most operators (I actually am unable to copy from the PDF, so I'm summarizing; basically everything from ternary `?:` to cast) should be made explicit.

***That said***, I should really change it to:

```
	return ::IsWow64Process(::GetCurrentProcess(), &f64) && f64;
```

...which looks less awkward anyways. Nothing can go wrong, since [the `return` statement beats everything](https://en.wikipedia.org/wiki/Comma_operator#Examples).

ariccio: (yes, I did change it)

UIforETW/Utility.h
ariccio: Whoops. Fixed.

UIforETW/Utility.h
ariccio: Fixed.

UIforETW/Utility.cpp
ariccio: You have a good point there. I think I'll remove a lot of it.

UIforETW/Utility.cpp
ariccio: ...pursuant to [a decision below](https://github.com/google/UIforETW/pull/62#discussion_r42307237), this is moot.

UIforETW/Utility.cpp
ariccio: Fixed.

UIforETW/Utility.cpp
ariccio: Fixed.

UIforETW/Utility.cpp
ariccio: Whoops, fixed. I've removed this function anyways.

ariccio:
There're a few stragglers, I'll address them tomorrow.

UIforETW/Utility.cpp
ariccio: Speaking of Chromium, in `scoped_handle.cc` there's a function called `CloseHandleWrapper` that `CHECK`s the return code.

This might actually be a good opportunity to try that `_Post_ptr_invalid_` trick of mine. (although I've grown a bit on `_Frees_ptr_`, which makes use of more advance annotations)

ariccio: (because `/analyze` sucks, you'd probably have to also annotate `HandleTraits::CloseHandle`, and then you'd get the full benefit)

ariccio: ..and there also appear to be a number of un-scoped-handle (raw) calls to `CloseHandle` in chromium:

https://code.google.com/p/chromium/codesearch#search/&q=CloseHandle%20case:yes%20exact:yes&m=100&sq=package:chromium&type=cs

...including a few in the sandbox?

UIforETW/Utility.cpp
randomascii: Another. It is better to have several small PRs rather than one enormous one.

UIforETW/Utility.cpp
randomascii: UIETWASSERT seems good. Are we actually leaking handles? If so it would be best to fix that.

UIforETW/Utility.cpp
randomascii: Absent evidence of a problem I think it's fine to do nothing. That is, if SetClipboardData fails then we're in terra incognita and any attempts to "handle" that are speculative at best, and harmful at worst. Unless the checks add clear value, why include them?

UIforETW/Utility.cpp
randomascii: Sure, if RegSetValueEx fails we might not set DisablePagingExecutive. And if a page-read fails then our code might not even get loaded and we'll crash. Both are possible, but both are unlikely, and in both cases we can't add any value by checking for them. For errors that should not happen the most that should be done is UIETWASSERT. Otherwise you're going down the rabbit hole.

UIforETW/Utility.cpp
randomascii: Cool.

I thought that the '=' statement was also extremely powerful - since it would be madness if it wasn't. Anyway, not a big deal, but it was sufficiently unusual to make it harder for me to read. And now it is zero deal.

UIforETW/Utility.cpp
randomascii: Thanks.

UIforETW/Utility.cpp
randomascii: There are more calls to CloseHandle than I would have guessed in Chromium. Curious. But maybe not that many for such a large code base?

While /analyze is good, I think that using ScopedHandle to make handle closing errors impossible is even better. And then the confusing annotations aren't needed. So, I'd stay away from the _Post_ptr_invalid_ trick (it will confuses readers and may not help /analyze) and stick with ScopedHandle style coding - clearly correct.

randomascii:
Let me know when you want me to look at it.

UIforETW/Utility.cpp
ariccio: Ok, I've actually replaced with an `ATLASSERT`, which requires much less code.

Using `UIETWASSERT` here necessitates:

```
	const LONG closeKey = ::RegCloseKey(key);
	UIETWASSERT(closeKey == ERROR_SUCCESS);
	(void)closeKey;//Debug only.
```

...whereas `ATLVERIFY` needs only:

```
	ATLVERIFY(::RegCloseKey(key) == ERROR_SUCCESS);
```

UIforETW/Utility.cpp
```diff
@@ -20,7 +20,152 @@ limitations under the License.
...
+HKEY OpenRegKey( _In_ const HKEY root, PCWSTR const subkey )
+{
+	HKEY key;
+	ATLVERIFY(::RegOpenKeyExW(root, subkey, 0, KEY_ALL_ACCESS, &key) == ERROR_SUCCESS);
+	return key;
+}
+
+void CloseRegKey(_In_ _Pre_valid_ _Post_ptr_invalid_ const HKEY key)
+{
+	ATLVERIFY(::RegCloseKey(key) == ERROR_SUCCESS);
```
ariccio: Note to self: Remove `CloseRegKey`

UIforETW/Utility.cpp
```diff
@@ -20,7 +20,152 @@ limitations under the License.
...
+}
+
+void CloseFindHandle(_In_ _Pre_valid_ _Post_ptr_invalid_ const HANDLE handle)
+{
+	ATLVERIFY(::FindClose(handle));
```
ariccio: Note to self: Remove `CloseFindHandle`

UIforETW/Utility.cpp
```diff
@@ -20,7 +20,152 @@ limitations under the License.
...
+	if (!unlockRes)
+	{
+		UIETWASSERT(::GetLastError() == NO_ERROR);
+	}
+}
+
+HKEY OpenRegKey( _In_ const HKEY root, PCWSTR const subkey )
+{
+	HKEY key;
+	ATLVERIFY(::RegOpenKeyExW(root, subkey, 0, KEY_ALL_ACCESS, &key) == ERROR_SUCCESS);
```
ariccio: Note to self: Remove `OpenRegKey`

UIforETW/Utility.cpp
randomascii: Good call.

ariccio: I'm actually going to remove the whole function. Now, there's only one line, so a function is just three lines of overhead.

ariccio:
Ok, I think I addressed everything. I don't think I'll be able to squash - 6 commits is annoying for sure.

It'd be nice if GitHub formatted the commit messages. Grr.

Here's the latest, for your convenience:

====
Code that I cut out was no longer necessary.

Drive by fixes:

1. Replaced manual COM memory management in GetDocumentsFolderPath with
[ATL's built-in COM smart pointer
(`ATL::CComHeapPtr`)](https://msdn.microsoft.com/en-us/library/80zw33a6.aspx).
I hate operator overloading, so I explicitly reference the `m_pData`
member of `ATL::CComHeapPtr`, which is otherwise implicitly passed to
`::SHGetKnownFolderPath` via the `operator&` overload.
2. I added `const` in a few places where it was missing. See
`ControlOK` (local variable `HWND parent`), `SetCurrentThreadName`
(local variable/argument to `::RaiseException`, `THREADNAME_INFO`).
3. I added parenthesis in a few places where multiple operators are used
in a single expression. See `GetNextDlgItem` (the `while` loop's
condition).
4. I added a comment to the call to `VirtualQuery` (with a link) that
documents the failure behavior, so it's clear that the `!=
sizeof(memoryInfo)` check is valid.

***NOTE: removing error checking causes `/analyze` to complain (after
all, we aren't correctly using APIs in the release build by not checking
their return values)***

====
Here are the warnings that `/analyze` generates:

```
C6102	utility.cpp
	286	Using 'key' from failed function call at line '285'.
	285	'key' is not initialized
	286	'key' is an Input to 'RegSetValueExW' (declared at c:\program files (x86)\windows kits\8.1\include\um\winreg.h:1004)
	286	'key' is used, but may not have been initialized

C6102	utility.cpp
	297	Using 'key' from failed function call at line '293'.
	293	'key' is not initialized
	297	'key' is an Input to 'RegCreateKeyW' (declared at c:\program files (x86)\windows kits\8.1\include\um\winreg.h:311)
	297	'key' is used, but may not have been initialized

C6387	utility.cpp
	584	Invalid parameter value	'ptr' could be '0':  this does not adhere to the specification for the function 'wcscpy_s'.
	581	'ptr' may be NULL
	584	'ptr' is an In/Out argument to 'wcscpy_s' (declared at c:\program files (x86)\microsoft visual studio 12.0\vc\include\string.h:291)
	584	'ptr' should not be NULL, because this is not consistent with the SAL annotation on 'wcscpy_s'
```

ariccio:
Hold up, hit an assertion here.

ariccio:
Ok, it's good now.

ariccio:
Ping?

randomascii:
I'll try to get to this today.

I'm going to try to figure out how to squash it to one commit. One guaranteed method is to have two repos and to xcopy the changed files from one to the other. It's weird that squash doesn't work that well.

ariccio:
It had something to do with merging changes from master, I think. I bet it wouldn't have been a problem if all UIforETW work was done on separate branches. That'd be impractical for such a small project, of course.

randomascii:
You're probably right. Regardless of the cause, the seven existing commits would make for weird history. The commit messages don't really tell the story of what this one change does, and why. Squashing it to a single commit with an explanatory commit message would make it more elegant.

I know how to do it (xcopy to a clean branch) and I'm just wondering if that would be worth you doing that, in order to have sensible history.

Looking at the code now...

UIforETW/Utility.cpp
```diff
@@ -349,78 +558,83 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
-	return result;
+	//TODO: IFileOperation?
+	return ::SHFileOperationW(&fileOp);
 }
 
 void SetClipboardText(const std::wstring& text)
 {
-	BOOL cb = OpenClipboard(GetDesktopWindow());
-	if (!cb)
+	ATLVERIFY(::OpenClipboard(::GetDesktopWindow()));
```
randomascii: Is OpenClipboard guaranteed to work? I don't think so, and now the code doesn't handle that case.

It still handles GlobalAlloc failing, which should actually never happen, so handling OpenClipboard seems like a good idea (while extremely rare, I think it's possible).

UIforETW/Utility.cpp
```diff
@@ -349,78 +558,83 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
+		return;
 	}
-
-	CloseClipboard();
+	ATLVERIFY(::CloseClipboard());
 }
 
 std::wstring GetClipboardText()
 {
+	ATLVERIFY(::OpenClipboard(::GetDesktopWindow()));
```
randomascii: Ditto on not handling OpenClipboard failing.

UIforETW/Utility.cpp
```diff
@@ -349,78 +558,83 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
 	std::wstring result;
-	BOOL cb = OpenClipboard(GetDesktopWindow());
-	if (!cb)
-		return result;
 
-	HANDLE hClip = GetClipboardData(CF_UNICODETEXT);
-	if (hClip)
+	HANDLE hClip = ::GetClipboardData(CF_UNICODETEXT);
```
randomascii: Woah - GetClipboardData can *definitely* fail. There might not be any text in the clipboard - could be an image instead. Possible errors still need to be handled.

UIforETW/Utility.cpp
```diff
@@ -349,78 +558,83 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
 }
 
 int64_t GetFileSize(const std::wstring& path)
 {
-	LARGE_INTEGER result;
-	HANDLE hFile = CreateFile(path.c_str(), GENERIC_READ,
-		FILE_SHARE_READ | FILE_SHARE_WRITE, NULL, OPEN_EXISTING,
+	LARGE_INTEGER result = {0};
+	HANDLE hFile = ::CreateFileW(path.c_str(), GENERIC_READ,
+		(FILE_SHARE_READ | FILE_SHARE_WRITE), NULL, OPEN_EXISTING,
```
randomascii: These parentheses seem excessive. Not a blocker, just sayin'.

UIforETW/Utility.cpp
```diff
@@ -547,9 +770,9 @@ std::wstring GetBuildTimeFromAddress(void* codeAddress)
 	// Print out the module information. The %.24s is necessary to trim
 	// the new line character off of the date string returned by asctime().
 	// _wasctime_s requires a 26-character buffer.
-	wchar_t ascTimeBuf[26];
+	wchar_t ascTimeBuf[26] = {0};
```
randomascii: I try to avoid "= {0}" - prefer "= {}". But, not a big deal.

I'm also curious as to why you initialize these at all? The _s functions guarantee null-termination, always, so this is provably unneeded waste.

UIforETW/Utility.h
```diff
@@ -39,11 +39,15 @@ void CreateRegistryKey(HKEY root, const std::wstring& subkey, const std::wstring
 std::wstring GetEditControlText(HWND hwnd);
 std::wstring AnsiToUnicode(const std::string& text);
 
+int RequiredNumberOfWideChars(const std::string& text);
```
randomascii: Is this needed outside of utility.cpp?

UIforETW/Utility.h
```diff
@@ -39,11 +39,15 @@ void CreateRegistryKey(HKEY root, const std::wstring& subkey, const std::wstring
 std::wstring GetEditControlText(HWND hwnd);
 std::wstring AnsiToUnicode(const std::string& text);
 
+int RequiredNumberOfWideChars(const std::string& text);
+
 // Return a string from a format string and some printf-style arguments.
 // Maximum output size is 4 K - larger outputs will be truncated.
-std::wstring stringPrintf(_Printf_format_string_ const wchar_t* pFormat, ...);
+std::wstring stringPrintf(_Printf_format_string_ PCWSTR const pFormat, ...);
```
randomascii: Note that the const on this function and the next is meaningless on a declaration. It only has meaning on the function definition.

randomascii:
Okay, I took a look. My one real concern is that there are three plays where error checking is removed that is actually needed!

With those fixed (or I can fix after merging the PR) it's all good. But, having it in one commit (as-is would be fine, just xcopy and recreate) would be better, with a great commit message.

randomascii:
When you do a "git pull" do you do use the --merge option? That tends to create cleaner histories that then work better with squash, I believe.

Anyway, let me know what you want to do. There are two questions:
1) Do you want to squash the commits (with another repo and an xcopy to get it down to one commit) or should I? I should be able to temporarily set my username/email address so that it shows in the history as you.
2) Do you want to add back in the error checking that is needed (clipboard copy in particular can definitely fail if there is no text in the clipboard) or should I?

It would be great to get this resolved soon.

ariccio:
I'll get back to you before the night's end. If I don't, go ahead.

UIforETW/Utility.cpp
ariccio: Big whoops there, that's definitely obvious in retrospect.

UIforETW/Utility.cpp
ariccio: I *figure* that if they do indeed null terminate, then the compiler will eliminate the zero initialization code during optimization, so it won't hurt. I should look at the actual generated code, though.

UIforETW/Utility.h
ariccio: Nope. Oops.

UIforETW/Utility.cpp
randomascii: That's expecting a lot. The CRT does the null termination and it would be pretty ballsy of the compiler to assume that of the CRT. It would have to read and interpret the /analyze annotations, but it actually ignores them during regular compiles. And that's good, because they are supposed to be analysis hints, not optimization hints.

UIforETW/Utility.cpp
mwinterb: Two basically irrelevant comments, but I want some of this useless knowledge to be useful:
1) From following the VsVim project, OpenClipboard can definitely fail in relatively normal scenarios.
2) GlobalAlloc(GMEM_MOVEABLE, ...) failing isn't quite as "should actually never happen" as with normal malloc -- there's a limit of about 64k outstanding moveable allocations. Given the expected frequency of that category of allocations, that's still well within the something is terribly wrong category if it fails, but it's not quite as dramatic as with normal address space exhaustion.

UIforETW/Utility.cpp
ariccio: Yup, indeed, the compiler does not optimize away.

UIforETW/Utility.cpp
randomascii: My guess is that each GMEM_MOVEABLE allocation has a 64 KB alignment requirement and therefore in a 32-bit app there can, logically, be no more than 64k of them outstanding. Same thing with VirtualAlloc(). But not a concern.

That is, I'm guessing that if you do 64k GMEM_MOVEABLE allocations then you have exhausted address space. But I don't know that for a fact.

ariccio:
I'm manually copying the files to a new branch, and reconstituting a commit message - to be opened in a new PR. It's probably easiest for both of us if you add the checking back in.

UIforETW/Utility.cpp
mwinterb: While reasonable, and address space exhaustion with low private byte consumption probably was not something I would have thought when I encountered it, it doesn't have the same "feel" as the analogous situation with VirtualAlloc: most notably that the same number pops up on a 64bit app. It seems like an explicit limit to balance implementation simplicity with a reasonable max for the remaining use cases for GMEM_MOVEABLE.

randomascii: There could be a 16-bit value in there somewhere, or perhaps they have to go in the bottom 4 GB.

Anyway, I'm not convinced that handling that type of memory exhaustion makes any more sense than handling malloc failing.

Where did you find documentation of the 64k limit? I don't see it?

mwinterb: That's fair, I was mainly just tossing it out as a note that GlobalAlloc failing wasn't necessarily in the same category as malloc failing.

The limit on count was found by experimentation. I first encountered it on XP or Vista at least and it is still there as of Win10. https://msdn.microsoft.com/en-us/library/ms810603.aspx <-- That references the limit, but it is an archived and ancient source. "If you work out the math you can see that the number of memory handles available for MOVEABLE memory is limited. 512K bytes / 8 bytes per handle = 65,535 handles. In fact, the system imposes this limitation on each process. Fortunately, this only applies to MOVEABLE memory."

randomascii: That makes sense. I guess they allocate/reserve a 512 KB block when you first call GlobalAlloc with GMEM_MOVEABLE and put all the handles in there. Having a hard limit lets them simplify the implementation. I could probably work out more details but it's not important. I would expect that there would never be more than a couple of outstanding allocations, barring leaks.

randomascii:
This was copied to pull request #64 which was then merged. Closing this one.

PR #64, opened 2015-10-27
Tweak utility funcs try five, redo of #62

ariccio:
Commit message below:
-----

Changes:
- Added lots of global scope qualifiers
- Added some const. See `ControlOK` (local variable `HWND parent`),
`SetCurrentThreadName` (local variable/argument to `::RaiseException`,
`THREADNAME_INFO`).
- Removed custom GetWindowsVersion, and replaced with
IsWindows*OrGreater variants
- Refactored CopyStartupProfiles
- Replaced `SHGetSpecialFolderPath` (depreciated) with
`SHGetKnownFolderPath`
- Added parenthesis for complex conditional expressions (see
`GetFileSize` and `GetNextDlgItem`)
- Added a comment to the call to `VirtualQuery` (with a link) that
documents the failure behavior, so it's clear that the `!=
sizeof(memoryInfo)` check is valid.
- Added a `static_assert` for the value of the third parameter to
`::RaiseException`, to statically check precondition.
- Replaced manual COM memory management in GetDocumentsFolderPath with
[ATL's built-in COM smart
pointer(`ATL::CComHeapPtr`)](https://msdn.microsoft.com/en-us/library/80zw33a6.aspx).
I hate operator overloading, so I explicitly reference the `m_pData`
member of `ATL::CComHeapPtr`, which is otherwise implicitly passed to
`::SHGetKnownFolderPath` via the `operator&` overload.

Details on the `SHGetSpecialFolderPath` depreciation:

SHGetSpecialFolderPath is NOT supported, as
(https://msdn.microsoft.com/en-us/library/windows/desktop/bb762204.aspx)
says:

```
[SHGetSpecialFolderPath is not supported. Instead, use ShGetFolderPath.]
```

... but SHGetFolderPath
(https://msdn.microsoft.com/en-us/library/windows/desktop/bb762181.aspx)
is *itself* depreciated, so I've refactored to use SHGetKnownFolderPath,
which is supported.

### DIFF
--- a/UIforETW/Utility.cpp
+++ b/UIforETW/Utility.cpp
@@ -20,7 +20,133 @@ limitations under the License.
 #include <direct.h>
 #include <ShlObj.h>
 
-std::vector<std::wstring> split(const std::wstring& s, char c)
+
+namespace {
+PCWSTR const kWPAStartupFileName = L"\\Startup.wpaProfile";
+
+void UnlockGlobalMemory(_In_ const HGLOBAL hmem)
+{
+	const BOOL unlockRes = ::GlobalUnlock(hmem);
+	if (!unlockRes)
+	{
+		UIETWASSERT(::GetLastError() == NO_ERROR);
+	}
+}
+
+std::wstring GetDocumentsFolderPath()
+{
+	ATL::CComHeapPtr<_Null_terminated_ wchar_t> docsPathTemp;
+	const HRESULT docsPathResult =
+		::SHGetKnownFolderPath(
+		FOLDERID_Documents, KF_FLAG_NO_ALIAS, NULL, &docsPathTemp.m_pData);
+	if (FAILED(docsPathResult))
+	{
+		debugPrintf(L"SHGetKnownFolderPath (for Documents) failed to retrieve the path.\n");
+		std::terminate();
+	}
+	return docsPathTemp.m_pData;
+}
+
+void copyWPAProfileToDocuments(const bool force)
+{
+	// First copy the WPA 8.1 startup.wpaProfile file
+	std::wstring docsPath(GetDocumentsFolderPath());
+
+	if (force)
+		outputPrintf(L"\n");
+
+	if (docsPath.empty())
+	{
+		outputPrintf( L"Failed to copy WPA profile to documents. See debugger output for details.\n" );
+		return;
+	}
+
+	const std::wstring source = docsPath + kWPAStartupFileName;
+	const std::wstring destDir = docsPath + std::wstring(L"\\WPA Files");
+	const std::wstring dest = destDir + kWPAStartupFileName;
+	const BOOL destinationExists = ::PathFileExistsW(dest.c_str());
+	if (force || !destinationExists)
+	{
+		ATLVERIFY(::CreateDirectoryW(destDir.c_str(), NULL));
+		const BOOL copyResult = ::CopyFileW(source.c_str(), dest.c_str(), FALSE);
+		if (copyResult)
+		{
+			if (force)
+				outputPrintf(L"Copied Startup.wpaProfile to the WPA Files directory.\n");
+			return;
+		}
+		if (force)
+		{
+			outputPrintf(L"Failed to copy Startup.wpaProfile to the WPA Files directory.\n");
+			outputLastError();
+			return;
+		}
+	}
+}
+
+void copyWPAProfileToLocalAppData(const std::wstring& exeDir, const bool force)
+{
+	PCWSTR const localAppDataEnvVar = L"localappdata";
+	// Then copy the WPA 10 startup.wpaProfile file
+	const std::wstring localAppData = GetEnvironmentVariableString(localAppDataEnvVar);
+	if (localAppData.empty())
+	{
+		outputPrintf(L"the `%s` environment variable didn't contain a valid path. Failed to copy WPA 10 profile.\n", localAppDataEnvVar);
+		return;
+	}
+	std::wstring source = exeDir + L"\\startup10.wpaProfile";
+	std::wstring destDir = std::wstring(localAppData) + L"\\Windows Performance Analyzer";
+	std::wstring dest = destDir + kWPAStartupFileName;
+	if (force || !::PathFileExistsW(dest.c_str()))
+	{
+
+		ATLVERIFY(::CreateDirectoryW(destDir.c_str(), NULL));
+
+		if (::CopyFileW(source.c_str(), dest.c_str(), FALSE))
+		{
+			if (force)
+				outputPrintf(L"%s", L"Copied Startup.10wpaProfile to %localappdata%\\Windows Performance Analyzer\n");
+			return;
+		}
+		if (force)
+		{
+			outputPrintf(L"%s", L"Failed to copy Startup.10wpaProfile to %localappdata%\\Windows Performance Analyzer\n");
+			outputLastError();
+		}
+	}
+
+}
+} //namespace {
+
+void outputLastError(const DWORD lastErr)
+{
+	const DWORD errMsgSize = 1024u;
+	wchar_t errBuff[errMsgSize] = {0};
+	const DWORD ret = ::FormatMessageW(
+		(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS),
+		NULL, lastErr, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+		errBuff, errMsgSize, NULL);
+
+	if (ret == 0)
+		std::terminate();//FormatMessageW failed.
+	outputPrintf(errBuff);
+}
+
+void debugLastError(const DWORD lastErr)
+{
+	const DWORD errMsgSize = 1024u;
+	wchar_t errBuff[errMsgSize] = {0};
+	const DWORD ret = ::FormatMessageW(
+		(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS),
+		NULL, lastErr, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+		errBuff, errMsgSize, NULL);
+
+	if (ret == 0)
+		std::terminate();//FormatMessageW failed.
+	debugPrintf(L"UIforETW encountered an error: %s\r\n", errBuff);
+}
+
+std::vector<std::wstring> split(const std::wstring& s, const char c)
 {
 	std::wstring::size_type i = 0;
 	std::wstring::size_type j = s.find(c);
@@ -40,26 +166,35 @@ std::vector<std::wstring> split(const std::wstring& s, char c)
 	return result;
 }
 
-std::vector<std::wstring> GetFileList(const std::wstring& pattern, bool fullPaths)
+std::vector<std::wstring> GetFileList(const std::wstring& pattern, const bool fullPaths)
 {
-	std::wstring directory;
-	if (fullPaths)
-		directory = GetDirPart(pattern);
+	const std::wstring directory = (fullPaths ? GetDirPart( pattern ) : L"");
+
+	//may not pass an empty string to FindFirstFileEx
+	UIETWASSERT(pattern.length() > 0);
+
+	//string passed to FindFirstFileEx must not end in a backslash
+	UIETWASSERT(pattern.back() != L'\\');
+
 	WIN32_FIND_DATA findData;
-	HANDLE hFindFile = FindFirstFileEx(pattern.c_str(), FindExInfoStandard,
+	HANDLE hFindFile = ::FindFirstFileExW(pattern.c_str(), FindExInfoBasic,
 				&findData, FindExSearchNameMatch, NULL, 0);
 
 	std::vector<std::wstring> result;
-	if (hFindFile != INVALID_HANDLE_VALUE)
+	if (hFindFile == INVALID_HANDLE_VALUE)
 	{
-		do
-		{
-			result.emplace_back(directory + findData.cFileName);
-		} while (FindNextFile(hFindFile, &findData));
-
-		FindClose(hFindFile);
+		//If there are NO matching files, then FindFirstFileExW returns
+		//INVALID_HANDLE_VALUE and the last error is ERROR_FILE_NOT_FOUND.
+		UIETWASSERT(::GetLastError() == ERROR_FILE_NOT_FOUND);
+		return result;
 	}
+	do
+	{
+		result.emplace_back(directory + findData.cFileName);
+	} while (::FindNextFileW(hFindFile, &findData));
 
+	UIETWASSERT(::GetLastError() == ERROR_NO_MORE_FILES);
+	ATLVERIFY(::FindClose(hFindFile));
 	return result;
 }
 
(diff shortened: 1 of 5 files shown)


randomascii:
Okay, merged. I'll push it upstream in a bit.

UIforETW/Utility.cpp
```diff
@@ -67,27 +202,42 @@ std::vector<std::wstring> GetFileList(const std::wstring& pattern, bool fullPath
...
 std::wstring LoadFileAsText(const std::wstring& fileName)
 {
+	if (!::PathFileExistsW(fileName.c_str()))
+		return L"";
+
 	std::ifstream f;
 	f.open(fileName, std::ios_base::binary);
 	if (!f)
+	{
+		outputPrintf(L"Failed to load file `%s` as text - f.open failed!\n", fileName.c_str());
```
randomascii: Error messages should be aimed at users, not developers. f.open failed seems rather low-level. TMI I would say.

UIforETW/Utility.cpp
```diff
@@ -127,71 +276,125 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
+void SetRegistryDWORD(const HKEY root, const std::wstring& subkey, const std::wstring& valueName, const DWORD value)
 {
 	HKEY key;
-	LONG result = RegOpenKeyEx(root, subkey.c_str(), 0, KEY_ALL_ACCESS, &key);
-	if (result == ERROR_SUCCESS)
-	{
-		RegSetValueEx(key, valueName.c_str(), 0, REG_DWORD, reinterpret_cast<const BYTE*>(&value), sizeof(value));
-		RegCloseKey(key);
-	}
+	ATLVERIFY(::RegOpenKeyExW(root, subkey.c_str(), 0, KEY_ALL_ACCESS, &key) == ERROR_SUCCESS);
```
randomascii: This changes RegOpenKeyEx from being a run-time error to being a logic error. I'm going to change it back because it's not a logic error.

UIforETW/Utility.cpp
```diff
@@ -127,71 +276,125 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
+	ATLVERIFY(ERROR_SUCCESS == ::RegSetValueExW(key, valueName.c_str(), 0, REG_DWORD, reinterpret_cast<const BYTE*>(&value), sizeof(value)));
```
randomascii: I'm leaving the ATLVERIFY on the other two calls, but I'm not convinced that they are worth it. There is definitely a readability cost, especially when the lines get up to 140 characters lon.

UIforETW/Utility.cpp
```diff
@@ -127,71 +276,125 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
-void CreateRegistryKey(HKEY root, const std::wstring& subkey, const std::wstring& newKey)
+void CreateRegistryKey(const HKEY root, const std::wstring& subkey, const std::wstring& newKey)
 {
 	HKEY key;
-	LONG result = RegOpenKeyEx(root, subkey.c_str(), 0, KEY_ALL_ACCESS, &key);
-	if (result == ERROR_SUCCESS)
+	ATLVERIFY(::RegOpenKeyExW(root, subkey.c_str(), 0, KEY_ALL_ACCESS, &key) == ERROR_SUCCESS);
+
+	HKEY resultKey;
+	//TODO: RegCreateKey is depreciated.
```
randomascii: Deprecated doesn't mean it's going away (it's not) and it doesn't mean we have to stop using it. It's just a hint that there might be better options. It doesn't mean they are necessarily worth switching to.

UIforETW/Utility.cpp
```diff
@@ -127,71 +276,125 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
+{
+	static_assert( sizeof(std::string::value_type) == 1 == sizeof(text[0]),
+		"bad assumptions!");
+
+	const int multiCharCount = ::MultiByteToWideChar(CP_ACP, 0, text.c_str(),
+		static_cast<int>(text.size() + 1), NULL, 0);
+
+	if (multiCharCount == 0)
+	{
+		//No reasonable way for MultiByteToWideChar to fail.
```
randomascii: Are you integration testing Windows?

UIforETW/Utility.cpp
```diff
@@ -20,7 +20,133 @@ limitations under the License.
...
+void outputLastError(const DWORD lastErr)
+{
+	const DWORD errMsgSize = 1024u;
+	wchar_t errBuff[errMsgSize] = {0};
+	const DWORD ret = ::FormatMessageW(
+		(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS),
+		NULL, lastErr, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+		errBuff, errMsgSize, NULL);
+
+	if (ret == 0)
```
randomascii: Why should the program terminate if there is a problem with printing an error message?

UIforETW/Utility.cpp
```diff
@@ -127,71 +276,125 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
+		//No reasonable way for MultiByteToWideChar to fail.
+		debugLastError( );
+		std::terminate( );
 	}
 
+	UIETWASSERT(multiToWideResult > 0);
+	UIETWASSERT(buffer[multiToWideResult - 1] == 0);
+
+	// Double-verify that the buffer is null-terminated.
+	buffer.at(buffer.size() - 1) = 0;
```
randomascii: Why buffer.at rather than buffer[]? And why the temporary 'result' variable? Don't just change code to match your preferred personal style.

UIforETW/Utility.cpp
```diff
@@ -255,22 +458,23 @@ static HWND GetNextDlgItem(HWND win, bool Wrap)
...
+void SmartEnableWindow(const HWND Win, const BOOL Enable)
 {
 	UIETWASSERT(Win);
 	if (!Enable)
 	{
-		HWND hasfocus = GetFocus();
 		bool FocusProblem = false;
-		HWND focuscopy;
-		for (focuscopy = hasfocus; focuscopy; focuscopy = (GetParent)(focuscopy))
+		for (HWND focuscopy = ::GetFocus(); focuscopy; focuscopy = ::GetParent(focuscopy))
```
randomascii: Good - reducing the scope of focuscopy is an improvement.

UIforETW/Utility.cpp
```diff
@@ -349,78 +558,83 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
 		FO_DELETE,
 		&fileNames[0],
 		NULL,
-		FOF_ALLOWUNDO | FOF_FILESONLY | FOF_NOCONFIRMATION,
+		(FOF_ALLOWUNDO | FOF_FILESONLY | FOF_NOCONFIRMATION),
```
randomascii: I think the extra parentheses are confusing. Removing.

UIforETW/Utility.cpp
```diff
@@ -349,78 +558,83 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
 
-	HANDLE hClip = GetClipboardData(CF_UNICODETEXT);
-	if (hClip)
+	HANDLE hClip = ::GetClipboardData(CF_UNICODETEXT);
+	UIETWASSERT(hClip);
+
+	void* const ptr = ::GlobalLock(hClip);
+	UIETWASSERT(ptr != NULL);
+	const size_t bytes = ::GlobalSize(hClip);
+	if (bytes == 0)
```
randomascii: Do we want to handle GlobalSize failing? What if the block size is zero? Won't the results of handling bytes==0 be identical to the results of doing nothing? i.e.; a zero-lenth string returned?

UIforETW/Utility.cpp
```diff
@@ -349,78 +558,83 @@ int DeleteFiles(HWND hwnd, const std::vector<std::wstring>& paths)
...
-	HANDLE hFile = CreateFile(path.c_str(), GENERIC_READ,
-		FILE_SHARE_READ | FILE_SHARE_WRITE, NULL, OPEN_EXISTING,
+	LARGE_INTEGER result = {0};
+	HANDLE hFile = ::CreateFileW(path.c_str(), GENERIC_READ,
+		(FILE_SHARE_READ | FILE_SHARE_WRITE), NULL, OPEN_EXISTING,
 		FILE_ATTRIBUTE_NORMAL, NULL);
+
 	if (hFile == INVALID_HANDLE_VALUE)
+	{
+		debugPrintf(L"Failed to get file size!\n");
```
randomascii: Remember that OutputDebugString prints globally, to any program (such as dbgview) that is listening. Therefore it is rude to print excessive messages. They may interfere with messages that a developer is trying to view from their own app.

UIforETW/Utility.cpp
```diff
@@ -431,108 +645,117 @@ bool Is64BitWindows()
...
 	BOOL f64 = FALSE;
-	bool bIsWin64 = IsWow64Process(GetCurrentProcess(), &f64) && f64;
-	return bIsWin64;
+	return ::IsWow64Process(::GetCurrentProcess(), &f64) && f64;
 #endif
 }
 
 bool Is64BitBuild()
 {
+	//when MSVC gets constexpr, this is a perfect candidate.
```
randomascii: I'm not sure that constexpr would improve this code since we don't actually need to declare a variable. But, the ideal solution is probably #if defined(_WIN64), matching Is64BitWindows() above.

UIforETW/Utility.cpp
```diff
@@ -431,108 +645,117 @@ bool Is64BitWindows()
...
 
-	return kWindowsVersionXP;
+bool IsWindowsVistaOrLesser()
+{
+	return !IsWindows7OrGreater();
 }
 
+std::wstring GetEnvironmentVariableString(_In_z_ PCWSTR const variable)
+{
+	const rsize_t bufferSize = 512u;
```
randomascii: Hold on. These handle environment variables with a maximum length of 512 characters? I don't know about your path, but my path is over 1,500 characters. My _NT_SYMBOL_PATH is also pretty long. This is a *serious* restriction.

UIforETW/Utility.cpp
```diff
@@ -431,108 +645,117 @@ bool Is64BitWindows()
...
+std::wstring GetBuildTimeFromAddress(_In_ const void* const codeAddress)
 {
 	// Get the base of the address reservation. This lets this
 	// function be passed any function or global variable address
 	// in a DLL or EXE.
-	MEMORY_BASIC_INFORMATION	memoryInfo;
-	if (VirtualQuery(codeAddress, &memoryInfo, sizeof(memoryInfo)) != sizeof(memoryInfo))
+
+	//https://msdn.microsoft.com/en-us/library/windows/desktop/aa366902.aspx
+	//If the function fails, the return value is zero.
```
randomascii: More specifically, the return value is the number of bytes copied. I want *all* requested bytes copied, which is why I compare against the size I passed in.

UIforETW/Utility.cpp
```diff
@@ -582,18 +805,18 @@ typedef struct tagTHREADNAME_INFO
...
-	info.dwThreadID = dwThreadID;
-	info.dwFlags = 0;
-
+	const DWORD dwThreadID = ::GetCurrentThreadId();
+	const THREADNAME_INFO info = { 0x1000, threadName, dwThreadID, 0 };
 	__try
 	{
-		RaiseException(MS_VC_EXCEPTION, 0, sizeof(info) / sizeof(ULONG_PTR), (ULONG_PTR*)&info);
+		const DWORD numArguments = sizeof(info) / sizeof(ULONG_PTR);
+		static_assert(
```
randomascii: It seems a bit excessive to test Microsoft's sample code for validity.

Remember, this is a tool, not a test framework.

UIforETW/Utility.h
```diff
@@ -39,11 +39,15 @@ void CreateRegistryKey(HKEY root, const std::wstring& subkey, const std::wstring
 std::wstring GetEditControlText(HWND hwnd);
 std::wstring AnsiToUnicode(const std::string& text);
 
+int RequiredNumberOfWideChars(const std::string& text);
```
randomascii: Not necessary to have this in the header.

UIforETW/Utility.h
```diff
@@ -39,11 +39,15 @@ void CreateRegistryKey(HKEY root, const std::wstring& subkey, const std::wstring
 std::wstring GetEditControlText(HWND hwnd);
 std::wstring AnsiToUnicode(const std::string& text);
 
+int RequiredNumberOfWideChars(const std::string& text);
+
 // Return a string from a format string and some printf-style arguments.
 // Maximum output size is 4 K - larger outputs will be truncated.
-std::wstring stringPrintf(_Printf_format_string_ const wchar_t* pFormat, ...);
+std::wstring stringPrintf(_Printf_format_string_ PCWSTR const pFormat, ...);
```
randomascii: The extra 'const' is superfluous in a declaration.

UIforETW/Utility.cpp
```diff
@@ -431,108 +645,117 @@ bool Is64BitWindows()
...
+	const rsize_t bufferSize = 512u;
+	wchar_t buffer[bufferSize] = {0};
+	rsize_t sizeRequired = 0u;
+	const errno_t result = _wgetenv_s(&sizeRequired, buffer, variable);
+	if (result == 0)
+		return buffer;
+	return L"";
+}
 
+std::string GetEnvironmentVariableString(_In_z_ PCSTR const variable)
```
randomascii: The narrow character version isn't needed.

UIforETW/Utility.cpp
randomascii: I've confirmed that this is not a theoretical problem. The Python scripts all fail on my machine because the path is too long. I'll fix.

UIforETW/Utility.cpp
```diff
@@ -40,26 +166,35 @@ std::vector<std::wstring> split(const std::wstring& s, char c)
...
 {
-	std::wstring directory;
-	if (fullPaths)
-		directory = GetDirPart(pattern);
+	const std::wstring directory = (fullPaths ? GetDirPart( pattern ) : L"");
+
+	//may not pass an empty string to FindFirstFileEx
+	UIETWASSERT(pattern.length() > 0);
+
+	//string passed to FindFirstFileEx must not end in a backslash
```
randomascii: Comments should have a space and then a capital letter. 'cause that's what the other comments have.

UIforETW/Utility.cpp
randomascii: More to the point, I hope you're right, since if it fails it will now take down UIforETW.

UIforETW/Utility.cpp
```diff
@@ -278,59 +482,64 @@ void SmartEnableWindow(HWND Win, BOOL Enable)
...
 std::wstring GetDirPart(const std::wstring& path)
 {
-	const wchar_t* pLastSlash = wcsrchr(path.c_str(), '\\');
-	if (pLastSlash)
-		return path.substr(0, pLastSlash + 1 - path.c_str());
+	UIETWASSERT(path.size() > 0);
+	const size_t lastSlash = path.find_last_of( L'\\' );
+	if (lastSlash != std::wstring::npos)
+	{
+		UIETWASSERT(path.at( lastSlash ) != path.back());
```
randomascii: This assert confuses me. First of all, the use of path.at()? Why?

Second, path.at( lastSlash ) will be equal to '\\'. So you're just checking to see if the last character is a '\\'. But that's not illegal, so it's not a valid assert.

UIforETW/Utility.cpp
```diff
@@ -127,71 +276,125 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
+//MultiByteToWideChar: https://msdn.microsoft.com/en-us/library/windows/desktop/dd319072.aspx
+//
+//Remarks:
+//As mentioned in the caution above,
+//the output buffer can easily be overrun
+//if this function is not first called with cchWideChar set to 0
+//in order to obtain the required size. 
+int RequiredNumberOfWideChars(const std::string& text)
+{
+	static_assert( sizeof(std::string::value_type) == 1 == sizeof(text[0]),
```
randomascii: This seems to be asserting that sizeof(char) is 1, which is true by definition. Harmless, mostly.

UIforETW/Utility.cpp
```diff
@@ -278,59 +482,64 @@ void SmartEnableWindow(HWND Win, BOOL Enable)
 
 std::wstring GetFilePart(const std::wstring& path)
 {
-	const wchar_t* pLastSlash = wcsrchr(path.c_str(), '\\');
-	if (pLastSlash)
-		return pLastSlash + 1;
+	UIETWASSERT(path.size() > 0);
+	const size_t lastSlash = path.find_last_of( L'\\' );
```
randomascii: Remember that L'\\' is the same as '\\', so that L is just decoration. This is true for all of the 7-bit ASCII characters.

UIforETW/Utility.cpp
ariccio: I agree, this should be fixed. Not sure why everything ran fine on my machine? My PATH is definitely more than 512 characters. I've had trouble with NSIS installers because of it.

UIforETW/Utility.cpp
ariccio: Well, Microsoft's sample code isn't always valid ;)

But yup, probably excessive.

ariccio:
I'll respond in more detail tomorrow. In brief: I agree with just about everything, and am glad to have this finally merged.

randomascii:
Be sure to check the follow-on commit.

Also, remember that 'deprecated' doesn't mean that we can't or shouldn't use. Deprecated Windows functions will *never* go away, and can be used as long as they still work.

randomascii:
Note that this change ended up having a few bugs. Some were fixed with the follow-up change 8cfb0d22. Another bug or two was fixed as part of change f635c725. Utility.cpp had two calls to CreateDirectoryW() that were wrapped with ATLVERIFY when in fact there was no reason to supposed that they would always succeed. Keeping PRs small makes it easier to ensure that all changes are tested, in debug and release builds. All newly written lines of code should be executed to make sure they behave as intended.

PR #65, opened 2015-10-27
Add a tool to generate a flame graph from an ETW trace.

fdoray:
A flame graph is a tool to visualize frequent call stacks efficiently.
Contrary to the existing flame graph functionality of UIforETW, this
tool generates a flame graph that includes off-CPU time. Knowledge of
Chrome is bundled into the tool to ignore uninteresting off-CPU time
(e.g. message loop waiting for a new task to execute).

Usage of the tool is described in ETWInsights\README.md.

Code to read an ETW trace that is not specific to flame graph
generation resides in the ETWInsights\src\etw_reader, as it could be
reused to build other tools (e.g. an ETW to chrome://tracing
converter).

### DIFF
--- a/ETWInsights/ETWInsights.sln
+++ b/ETWInsights/ETWInsights.sln
@@ -1,0 +1,41 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 2013
+VisualStudioVersion = 12.0.31101.0
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "base", "base\base.vcxproj", "{637388CA-E6E9-4D38-8DC9-CC2FE937DE24}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "etw_reader", "etw_reader\etw_reader.vcxproj", "{E1FCFE0C-B8CB-4516-9F46-54C1F73A9601}"
+	ProjectSection(ProjectDependencies) = postProject
+		{637388CA-E6E9-4D38-8DC9-CC2FE937DE24} = {637388CA-E6E9-4D38-8DC9-CC2FE937DE24}
+	EndProjectSection
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "flame_graph", "flame_graph\flame_graph.vcxproj", "{EE253535-588A-4ABC-A65E-8DE45E240D7F}"
+	ProjectSection(ProjectDependencies) = postProject
+		{E1FCFE0C-B8CB-4516-9F46-54C1F73A9601} = {E1FCFE0C-B8CB-4516-9F46-54C1F73A9601}
+		{637388CA-E6E9-4D38-8DC9-CC2FE937DE24} = {637388CA-E6E9-4D38-8DC9-CC2FE937DE24}
+	EndProjectSection
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Win32 = Debug|Win32
+		Release|Win32 = Release|Win32
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{637388CA-E6E9-4D38-8DC9-CC2FE937DE24}.Debug|Win32.ActiveCfg = Debug|Win32
+		{637388CA-E6E9-4D38-8DC9-CC2FE937DE24}.Debug|Win32.Build.0 = Debug|Win32
+		{637388CA-E6E9-4D38-8DC9-CC2FE937DE24}.Release|Win32.ActiveCfg = Release|Win32
+		{637388CA-E6E9-4D38-8DC9-CC2FE937DE24}.Release|Win32.Build.0 = Release|Win32
+		{E1FCFE0C-B8CB-4516-9F46-54C1F73A9601}.Debug|Win32.ActiveCfg = Debug|Win32
+		{E1FCFE0C-B8CB-4516-9F46-54C1F73A9601}.Debug|Win32.Build.0 = Debug|Win32
+		{E1FCFE0C-B8CB-4516-9F46-54C1F73A9601}.Release|Win32.ActiveCfg = Release|Win32
+		{E1FCFE0C-B8CB-4516-9F46-54C1F73A9601}.Release|Win32.Build.0 = Release|Win32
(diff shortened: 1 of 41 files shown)


ETWInsights/base.vcxproj
```diff
@@ -0,0 +1,220 @@
...
+      <TargetExt Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">.lib</TargetExt>
+      <OutDir Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">C:\src\UIforETW\ETWInsights\Release\</OutDir>
+      <IntDir Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">base.dir\Release\</IntDir>
+      <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">base</TargetName>
+      <TargetExt Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">.lib</TargetExt>
+      <OutDir Condition="'$(Configuration)|$(Platform)'=='MinSizeRel|Win32'">C:\src\UIforETW\ETWInsights\MinSizeRel\</OutDir>
+      <IntDir Condition="'$(Configuration)|$(Platform)'=='MinSizeRel|Win32'">base.dir\MinSizeRel\</IntDir>
+      <TargetName Condition="'$(Configuration)|$(Platform)'=='MinSizeRel|Win32'">base</TargetName>
+      <TargetExt Condition="'$(Configuration)|$(Platform)'=='MinSizeRel|Win32'">.lib</TargetExt>
+      <OutDir Condition="'$(Configuration)|$(Platform)'=='RelWithDebInfo|Win32'">C:\src\UIforETW\ETWInsights\RelWithDebInfo\</OutDir>
```
randomascii: Oops - the project files have hard-coded paths in them. I guess this is what CMake does, which means that it's project files are non-portable. Choices are:
1) Manually fix the paths to make them all relative.
2) Find the CMake switch to generate relative paths.
3) Recreate the project files using the project wizard.

I'm inclined to prefer option three because I think CMake projects are ugly, but this is sufficiently isolated that it's fine either way.

ETWInsights/etw_reader.vcxproj
```diff
@@ -0,0 +1,214 @@
...
+    <ClInclude Include="C:\src\UIforETW\ETWInsights\src\etw_reader\stack.h" />
+    <ClInclude Include="C:\src\UIforETW\ETWInsights\src\etw_reader\system_history.h" />
+    <ClInclude Include="C:\src\UIforETW\ETWInsights\src\etw_reader\thread_history.h" />
+    <ClCompile Include="C:\src\UIforETW\ETWInsights\src\etw_reader\etw_reader.cc"  />
+    <ClCompile Include="C:\src\UIforETW\ETWInsights\src\etw_reader\generate_history_from_trace.cc"  />
+    <ClCompile Include="C:\src\UIforETW\ETWInsights\src\etw_reader\system_history.cc"  />
+    <ClCompile Include="C:\src\UIforETW\ETWInsights\src\etw_reader\thread_history.cc"  />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="C:/src/UIforETW/ETWInsights/base.vcxproj">
```
randomascii: Project references are also using absolute paths. I only discovered this and the other absolute paths because my UIforETW directory is in a different location.

randomascii:
Once it's landed and building I can integrate it so that UIforETW can invoke it easily, and add it to the etw_package.bat build steps.

Does it work well with the unmodified flamegraph.pl from Brendan's github repo?

fdoray:
I recreated the project using the visual studio project wizard.

It would certainly be nice to integrate the tool with UIforETW. Would we ask the user for which process to generate the flame graph?

The output of the tool works well with the unmodified flamegraph.pl from Brendan's github repo. With my modified version of the script, stack frames from Chrome and Windows DLLs have different colors, which is nice but not required.

randomascii:
Sorry for the delay. Lots of other stuff. I'll look at this tonight or tomorrow morning.

I think that UIforETW would put up a dialog and let the user select a PID, possibly TID, and a time range.

If the different colors are nice enough we can always ship the patch file, or convince Brendan to incorporate it.

fdoray:
Sorry, I forgot to address your comment about reusing child_process.cc/.h from UIforETW. I'll change that tomorrow.

fdoray:
Hmmm... reusing your ChildProcess is not that simple. I would also have to reuse your Utility.h/.cpp files and your precompiled headers.

Knowing that even though I initially copied your ChildProcess implementation, they are now very different (yours writes output to a pipe while mine writes output to a file) + it's not that easy to add a reference to your files (Utility.h/.cpp + precompiled headers), is it OK if I keep my own implementation in ETWInsights/?

randomascii:
> is it OK if I keep my own implementation in ETWInsights

Yep, that's fine.

randomascii:
I have some suggestions:
- Should print all command-line options if none are specified
- Should print name of output file when done
- Should probably print reassuring status messages for different stages? Maybe?
- Should probably set the title to indicate what the flame graph is displaying (especially when run with a trace as the only parameter)
- The line "http://www.apache.org/licenses/LICENSE-2.0" got non-indented by auto-paste-formatting in some files, which makes it inconsistent.
- In ConvertEtlToCsv the WaitForCompletion call should be unneeded since the ChildProcess destructor will do that.
I also found one apparent bug:
- ERROR(etw_reader.cc:135): Unexpected number of tokens for line of type FileVersion.
And I have some changes I could make
- Build settings changes for faster builds/better warnings

But, it works, and none of these are blockers. So, I'm tempted to take the pull request and then we can iterate as needed in the main repo. Any thoughts?

fdoray:
I addressed these comments:

- Should print name of output file when done
- Should probably print reassuring status messages for different stages? Maybe?
- The line "http://www.apache.org/licenses/LICENSE-2.0" got non-indented by auto-paste-formatting in some files, which makes it inconsistent.
- In ConvertEtlToCsv the WaitForCompletion call should be unneeded since the ChildProcess destructor will do that. I also found one apparent bug:

Can I keep these comments for later, and merge this pull request now?

- Should print all command-line options if none are specified
- Should probably set the title to indicate what the flame graph is displaying (especially when run with a trace as the only parameter)  [I changed the name of the output file to {trace file}.flamegraph.txt, is that what you meant?]
- ERROR(etw_reader.cc:135): Unexpected number of tokens for line of type FileVersion. And I have some changes I could make [I'm curious about that. Could I see the trace file that generated this error?]
- Build settings changes for faster builds/better warnings

Thanks for the review.

randomascii:
Go for it. It works and we can iterate on it within version control.

Let me know if you are going to add UI to UIforETW to call it, otherwise I might do it this weekend.

fdoray:
I don't plan to work on adding UI for UIforETW before next week. You can do it.

PR #66, opened 2015-11-02
Show flame_graph.exe usage when incorrect arguments are specified.

fdoray:
With this CL, a short description of the arguments of
flame_graph.exe is shown when no arguments or invalid arguments
are specified.


randomascii:
lgtm

In a future change it could be helpful to have an option to take a PID, since process name is hopelessly vague when looking at Chrome. The tid is probably the more specifier that will be used more commonly but I could see having a pid being useful.

fdoray:
ok, I'll add pid in a future patch.

PR #67, opened 2015-11-04
Add Chrome regions of interest to WPA startup profiles.

fdoray:
This adds an XML file that tells WPA how to generate regions of
interest from Chrome ETW events. Also, WPA startup profiles are
updated so that this XML file is loaded by default.

### DIFF
--- a/UIforETW/Utility.cpp
+++ b/UIforETW/Utility.cpp
@@ -22,7 +22,8 @@ limitations under the License.
 
 
 namespace {
-PCWSTR const kWPAStartupFileName = L"\\Startup.wpaProfile";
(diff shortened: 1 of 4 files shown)


randomascii:
Is this the change you had to make to avoid the crash:

ATLVERIFY(::CreateDirectoryW(destDir.c_str(), NULL) || ::GetLastError() == ERROR_ALREADY_EXISTS);

Trivia: VS 2015 debug builds reset last error when you allocate or free memory, so I may have broken a few fragile uses of GetLastError() when I upgraded, possibly including this.

randomascii:
lgtm. Thanks for the fix of the two CreateDirectoryW calls. I can merge it or you can.

PR #73, opened 2016-03-24
Source compatibility with clang.

mwinterb:
This basically just addresses errors and the more 'silly' warnings: dllimport has no effect when defining a typedef and fully qualifying a class member function with its class name is unnecessary and a Microsoft-specific extension when declaring it.


randomascii:
Simple enough. Thanks for the fixes.

PR #74, opened 2016-03-27
add CLR runtime tracing support

MikeMarcin:
Add option to record symbols for ngen CLR dlls.

Allows me to see symbols for WindowsBase.ni.dll and PresentationFramework.ni.dll.

### DIFF
--- a/UIforETW/Support.cpp
+++ b/UIforETW/Support.cpp
@@ -62,6 +62,7 @@ void CUIforETWDlg::TransferSettings(bool saving)
 		{ L"SampledStacks", &bSampledStacks_ },
 		{ L"FastSampling", &bFastSampling_ },
 		{ L"GPUTracing", &bGPUTracing_ },
+		{ L"CLRTracing", &bCLRTracing_ },
 		{ L"ShowCommands", &bShowCommands_ },
 		{ L"ChromeDeveloper", &bChromeDeveloper_ },
 		{ L"AutoViewTraces", &bAutoViewTraces_ },
--- a/UIforETW/UIforETW.rc
+++ b/UIforETW/UIforETW.rc
@@ -74,6 +74,7 @@ END
 // remains consistent on all systems.
 IDR_MAINFRAME           ICON                    "res\\UIforETW.ico"
 
+
 /////////////////////////////////////////////////////////////////////////////
 //
 // Dialog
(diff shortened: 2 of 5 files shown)


UIforETW/UIforETWDlg.cpp
```diff
@@ -769,7 +774,7 @@ void CUIforETWDlg::OnBnClickedStarttracing()
 	else
 		UIETWASSERT(0);
 
-	std::wstring kernelProviders = L" Latency+POWER+DISPATCHER+DISK_IO_INIT+FILE_IO+FILE_IO_INIT+VIRT_ALLOC+MEMINFO";
+	std::wstring kernelProviders = L" Base+Latency+POWER+DISPATCHER+DISK_IO_INIT+FILE_IO+FILE_IO_INIT+VIRT_ALLOC+MEMINFO";
```
randomascii: Having Base+Latency is odd because they are both Kernel Groups and they overlap significantly. Base adds MEMINFO which is already explicitly included, and MEMINFO_WS which I have not found to be useful. If MEMINFO_WS is needed then make sure it works with the public version of WPA - I think the display of its data might be internal Microsoft only.

UIforETW/UIforETW.rc
```diff
@@ -298,6 +300,17 @@ END
 
 /////////////////////////////////////////////////////////////////////////////
 //
+// AFX_DIALOG_LAYOUT
+//
+
+IDD_UIFORETW_DIALOG AFX_DIALOG_LAYOUT
```
randomascii: What's this section? Accidentally added blank dialog?

UIforETW/Support.cpp
```diff
@@ -62,6 +62,7 @@ void CUIforETWDlg::TransferSettings(bool saving)
 		{ L"SampledStacks", &bSampledStacks_ },
 		{ L"FastSampling", &bFastSampling_ },
 		{ L"GPUTracing", &bGPUTracing_ },
+        { L"CLRTracing", &bCLRTracing_ },
```
randomascii: It looks like you edited the file using spaces and the rest of it uses tabs, so the lines don't line up when github displays them with an odd tab size. This repeats in quite a few places.

randomascii:
This looks good and looks like it will be a valuable addition. Do you have any information on what this does to trace sizes? On my one test it seemed to create a huge trace with lots of events lost.

The change description needs a title line followed by a blank line. git convention says that the title line should be no more than about 50 characters: http://chris.beams.io/posts/git-commit/

The mixture of present and past tense made it hard to parse what this actually does. For me these changes read more clearly (even if you are still having these cpu spikes):

< I'm getting cpu spikes
> I was getting cpu spikes
< It traces down
> It traced down

I think it's worth adding a sentence saying what this change actually does. i.e.; state explicitly that this gives symbol names to managed code call stacks. It could be as simple as "with this change I now get symbols for WindowsBase.ni.dll and PresentationFrammwork.ni.dll." If you have any information about what this does and doesn't capture (i.e.; what CLR rundown is supposed to add) then that would be helpful.

Spelling error in PresentationFrammwork.ni.dll in the comment.

randomascii:
Also relevant:

https://github.com/google/UIforETW/issues/48
https://github.com/google/UIforETW/pull/51

UIforETW/UIforETW.rc
mwinterb: It's related to the new dynamic layout support in MFC for VS2015:
https://msdn.microsoft.com/en-us/library/mt270148.aspx
Probably just got added by default since the dialog was modified.

randomascii: Interesting. Well, no harm - leave it or delete it.

Good to see Microsoft finally supporting sizeable dialogs - I've been implementing them for 20 years without any help, but maybe they'll become more common now.

UIforETW/UIforETWDlg.cpp
MikeMarcin: Added it because it was in the example:
https://msdn.microsoft.com/en-us/library/windows/hardware/hh448186.aspx
I still get symbols without it so it seems to be unnecessary.

UIforETW/Support.cpp
MikeMarcin: okay

MikeMarcin:
Looks like #51 adds the same support and also tries to add the Rundown Provider.
I think the Rundown should be a separate option as the documentation says:
> The rundown provider must be turned on for certain special-purpose uses. However, for a majority of users, the runtime provider should suffice.

Wish I'd known about #51 before I did this could've saved me some time.
My work is a subset of that PR so feel free to drop this.

randomascii:
I figured the 'Base' was probably unneeded - good to get that cleaned up.

#51 never got finished and it sounds like a subset might be better anyway. I'm happy to land your change - I'm sure I'll use it sometimes. Can you squash it to a single commit?

MikeMarcin:
After fighting with git for a bit I think I've managed to squash it.

randomascii:
Hmmm - I merged this but github isn't showing it as merged. And due to the questionable timing of my other change the history didn't end up as clean as I'd hoped. Oh well - it's in there! Thanks!

MagicAndre1981:
the rundown provider is required to capture _DC events, so data that occurred before the trace was started. Vance explained it here: https://channel9.msdn.com/Shows/Defrag-Tools/Defrag-Tools-116-PerfView-Part-4#time=19m10s

PR #82, opened 2016-04-28
Make flamegraph generation python script more flexible

surki:


surki:
I signed it!

randomascii:
Looks good! Can you add yourself to the AUTHORS and CONTRIBUTORS files and update this pull request with those changes? If you do it with --amend then the change will add most easily to the master branch. A more detailed description would also be nice, but isn't critical.

Thanks for contributing. And thanks googlebot for dealing with the CLA details.

surki:
Updated. Thanks.

randomascii:
Thanks for the change. I verified that it still works from the trace context menu. Now, if anyone wants to add some UI that will hook up the newly available options, that would be great.

Although, I hear rumors of flame graph support in future versions of WPA. Fingers crossed...

PR #87, opened 2016-08-26
Add minimize and maximum buttons

ricnewton:
Add minimize and maximum buttons so we can minimize the UI out the way once tracing started


ricnewton:
I signed it!

randomascii:
Merged! I'm not sure why it's not showing up, but it is merged in.

PR #88, opened 2016-08-27
Replace default MFC icon with new one.

ricnewton:
So a friend of mine who does UX stuff saw the default MFC icon and couldn't stop himself from replacing it :).

I think it is an accident that it looks a bit like the chrome one.


randomascii:
I knew that my strategy of having an intentionally ugly icon would eventually trigger an open-source response.

Thanks!

Now I just need a sound designer to create sounds for starting tracing, stopping tracing, starting recording, and finishing recording...

PR #91, opened 2016-10-13
Add currently pressed meta keys to the key name in KeyDown event

ricnewton:
Add currently pressed meta keys to the key name in KeyDown event so we can filter and highlight on keyboard combinations as discussed in https://github.com/google/UIforETW/pull/90.

I did not add a new field in the pull request but just extended what is put in field2 (key name) to keep it simple, I can add a new field if this would be preferred.

### DIFF
--- a/UIforETW/KeyLoggerThread.cpp
+++ b/UIforETW/KeyLoggerThread.cpp
@@ -25,6 +25,20 @@ namespace
 
 std::atomic<bool> g_LogKeyboardDetails(false);
 
+std::string MetaKeys()
+{
+	std::string metaKeys;
+	if (GetAsyncKeyState(VK_CONTROL))
+		metaKeys += "Ctrl+";
+	if (GetAsyncKeyState(VK_SHIFT))
+		metaKeys += "Shift+";
+	if (GetAsyncKeyState(VK_MENU))
+		metaKeys += "Alt+";
+	if (GetAsyncKeyState(VK_LWIN) || GetAsyncKeyState(VK_RWIN))
+		metaKeys += "Win+";
+	return metaKeys;
+}
+
 _Pre_satisfies_(nCode == HC_ACTION)
 LRESULT CALLBACK LowLevelKeyboardHook(int nCode, WPARAM wParam, LPARAM lParam)
 {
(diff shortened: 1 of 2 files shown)


randomascii:
I'll double-check tonight but after a quick examination it looks good. Retaining the current fields is slightly simpler, and I can't really think of any disadvantages, but I'll experiment with that to see how it feels.

ricnewton:
Cool, its working well for my workflow :).  I've update by starting WPA profile with a few chosen hotkeys and can zoom to the mark region instantly.

randomascii:
Maybe share the startup profile and describe your zoom workflow - sounds interesting/good.

ricnewton:
OK, put a basic UserMark.wpa profile in, the top trace is the keyboard shortcut, set to Ctrl+Alt+Win+F[1-12], then there is a CPU usage (precise) graph and a sampled table (on Windows 10 I also have a flame graph).

My workflow is to hit Ctrl+Alt+Win+F12, do whatever activity I want to profile, hit Ctrl+Alt+Win+F12 again then in WPA just choose Ctrl+Alt+Win+F12 on the top graph and then zoom the CPU graph to that range.

randomascii:
Okay, done!

It doesn't show up as merged because I tweaked the commit message of your second commit - subject line should be ~50 characters max, then a blank line, then arbitrary paragraphs. Otherwise "git log --oneline" works poorly. Otherwise I took it as is.

I find Ctrl+Alt+Win+F12 a hell of a keyboard-chord - I would have gone for something simpler - but that's just a .wpaProfile tweak (could be more permissive and allow all modified function keys to show up?) and matter of opinion. Overall I like the idea. It does make selecting time ranges easier, and it makes examining keyboard input (especially when switched to "Full" input tracing) that little bit easier.

There's a slightly weird disconnect between the names used for the special keys (control) and the names used for them as modifiers (Ctrl) but I couldn't decide if that was good or bad.

Anyway, thanks for the contribution.

ricnewton:
Great, thank.  Ctl+Alt+Win was just really easy on my keyboard, and I picked a function key to hopefully not clash with any app hotkey, I assumed people would teak it to whatever they liked.

I probably should have reused the control naming scheme, I've just always seen shortcut written as ctrl.

Thanks again

PR #100, opened 2017-03-02
Added support for TraceLogging providers

mwinterb:
Only strictly necessary for pre-Windows 8 operating systems as the
xperf from the Windows 10 Anniversary Edition version (August 2016)
SDK supports the proper translation "out of the box".

The algorithm to translate from TraceLogging provider name to GUID is
based on the C# code posted by Doug Cook to
https://blogs.msdn.microsoft.com/dcook/2015/09/08/etw-provider-names-and-guids/

### DIFF
--- a/UIforETW/TraceLoggingSupport.cpp
+++ b/UIforETW/TraceLoggingSupport.cpp
@@ -1,0 +1,231 @@
+/*
+Copyright 2017 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "stdafx.h"
+#include "TraceLoggingSupport.h"
+#include "Utility.h"
+
+#include <bcrypt.h>
+
+#include <vector>
+#include <array>
+
+#ifdef _DEBUG
+#define new DEBUG_NEW
+#endif
+
+#pragma comment(lib, "Bcrypt.lib")
+
+static std::system_error MakeSystemErrorFromWin32Error(DWORD errorCode, const char* context)
+{
+	std::error_code code(errorCode, std::system_category());
+	return std::system_error(code, context);
+}
+namespace
+{
+class SHA1HashProvider
+{
+	struct secret_init
+	{
+	};
+	// this is just to force the destructor to be called
+	// after this constructor finishes, regardless of whether or not
+	// the calling constructor throws an exception
+	SHA1HashProvider(secret_init)
+		: algHandle_(nullptr)
+		, hashHandle_(nullptr)
+	{
+	}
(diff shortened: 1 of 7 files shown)


mwinterb:
This is partly a WIP to solicit feedback. I've tested it with my two TraceLogging providers with extra options and it's functional, but if there are objections to the general path I'd rather find out sooner.

UIforETW/TraceLoggingSupport.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include <vector>
+#include <array>
+
+#ifdef _DEBUG
+#define new DEBUG_NEW
+#endif
+
+#pragma comment(lib, "Bcrypt.lib")
+
+static std::system_error MakeSystemErrorFromWin32Error(DWORD errorCode, const char* context)
```
mwinterb: One of the biggest "open questions" is that no one catches these exceptions, but the context for these exceptions mostly seem in the realm of out-of-memory only. Well, I guess I could handle them in the Starttracing handler and treat it as if Xperf failed to launch?

randomascii: Where are these exceptions documented? It's not possible for xperf.exe to throw an exception back to UIforETW.

randomascii:
Roughly speaking this seems good. I may not get a chance to try the code until next week.

If the change is squashed that is better - having it as a single commit is nice - but I doubt that is crucial.

UIforETW/TraceLoggingSupport.cpp
mwinterb: I added an exception handler in the Starttracing. I can see LCMapStringEx failing for some invalid strings, but I wasn't able to trivially cause that, and if MFC is going to catch that exception silently, then that would defeat the goal of the fail fast.

I also squashed the commits.

Mildly on topic, do you know if this issue https://randomascii.wordpress.com/2012/07/05/when-even-crashing-doesnt-work/ still applies to recent Windows versions?

randomascii:
Much to my surprise it looks like the kernel-callback-exception-swallowing has been fixed. Crazy. I wouldn't have thought Microsoft would be brave enough to do that.

randomascii:
Just a few tidiness things, if you don't mind. The issues are most obvious with "git show" on your change. The README change has a trailing space on one line (shows read on the git show) and the TraceLoggingSupport.* files have \n endings instead of \r\n (VS's regex replace can fix that). I can't test it easily but it looks plausible.

randomascii:
Uggh. I don't understand github. Or git for that matter. I merged your change in. I used a cherry-pick which has worked previously but which I'm probably being punished for. So naturally it thinks that it's a totally different change and that there is a conflict. Thinking it is a totally different change is probably because I committed a few changes after pulling yours into a branch. The conflicts are because I made some whitespace changes after cherry-picking yours.

But, it's all good. Your change is in. It seems to work? But give it a try and let me know. It will be in the next release.

And thanks!

mwinterb:
Sorry for not addressing this before you had to deal with that. On "my end" they seemed to all be CRLF, but git show did point out the problems. Since I didn't know what was wrong, I was going to look at it some evening, but my last two nights were "booked". I should have mentioned that rather than just being silent.

But: testing with "my" two TraceLogging providers captures the right data when specifying them in the settings dialog. Thanks for the merge and fixing up those problems!

randomascii:
No worries. It was trivial, just confusing when I saw the conflicts. But the change shows up with your name on it, and it sounds like it works, so that's all that matters.

PR #108, opened 2017-08-07
Minimize to system tray

aerworker:
The main window shows an icon in system tray when minimized and hides itself in task bar. Click on icon restores the main window.
Does it need to introduce an option for this functionality?

### DIFF
--- a/UIforETW/UIforETWDlg.cpp
+++ b/UIforETW/UIforETWDlg.cpp
@@ -172,6 +172,7 @@ void CUIforETWDlg::ShutdownTasks()
 		SetSamplingSpeed();
 		bFastSampling_ = true;
 	}
+	Shell_NotifyIcon(NIM_DELETE, &trayData);
 }
 
 void CUIforETWDlg::OnCancel()
@@ -261,6 +262,7 @@ BEGIN_MESSAGE_MAP(CUIforETWDlg, CDialog)
 	ON_WM_ACTIVATE()
 	ON_WM_TIMER()
 	ON_BN_CLICKED( IDC_CLRTRACING, &CUIforETWDlg::OnBnClickedClrtracing )
+	ON_MESSAGE(WM_TRAY_MESSAGE, &CUIforETWDlg::OnTrayNotify)
 END_MESSAGE_MAP()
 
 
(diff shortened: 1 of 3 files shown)


randomascii:
Can you explain why this is a good idea, preferably as part of the commit message? I'm a believer in more detailed commit messages - see this post for some good guidelines:

https://chris.beams.io/posts/git-commit/

But, I can't look at the patch until the CLA is signed. Thanks!

aerworker:
I signed it!

aerworker:
I've updated the commit.

UIforETW/UIforETWDlg.cpp
```diff
@@ -2409,3 +2416,59 @@ void CUIforETWDlg::OnActivate(UINT nState, CWnd* pWndOther, BOOL bMinimized)
...
+
+	trayData.uFlags = NIF_ICON | NIF_MESSAGE | NIF_TIP;
+	
+
+	BOOL bSuccess = FALSE;
+
+	bSuccess = Shell_NotifyIcon(NIM_ADD, &trayData);
+
+	if (!(bSuccess))
+		AfxMessageBox(L"Unable to Set Tary Icon");
```
randomascii: Do we really want to "handle" this error by bringing up a message box? Also, type in Tary/Tray

UIforETW/UIforETWDlg.cpp
```diff
@@ -2409,3 +2416,59 @@ void CUIforETWDlg::OnActivate(UINT nState, CWnd* pWndOther, BOOL bMinimized)
...
+		return 0;
+
+	if (uMsg == WM_LBUTTONDOWN) 
+	{
+		BOOL bSuccess = FALSE;
+
+		bSuccess = Shell_NotifyIcon(NIM_DELETE, &trayData);
+		
+		if (!(bSuccess))
+			AfxMessageBox(L"Unable to Delete Tary Icon");
```
randomascii: Ditto

UIforETW/UIforETWDlg.cpp
```diff
@@ -2409,3 +2416,59 @@ void CUIforETWDlg::OnActivate(UINT nState, CWnd* pWndOther, BOOL bMinimized)
...
+}
+
+LRESULT CUIforETWDlg::OnTrayNotify(WPARAM wParam, LPARAM lParam)
+{
+	UINT uID;
+	UINT uMsg;
+
+	uID = (UINT)wParam;
+	uMsg = (UINT)lParam;
+
```
randomascii: Avoid double blank lines

randomascii:
So... I'm not seeing an updated commit message with an explanation of why, and I definitely have concerns. This seems like unusual behavior, and confusing.

Currently I can tell whether UIforETW is running by looking in one place for its icon (especially since I have it pinned). With this change it might be there, or it might be in the notification area, and the notification might be hidden.

Now, if this was hidden behind a switch then this would not be as severe a problem. But that's more code, more complexity in the settings dialog, more ways for users to get themselves into a confused state.

This doesn't necessarily mean "no", but it does mean "convince me of why this is good". Precedent for this UI pattern would be helpful.

Issue #118, opened 2018-05-02
UIforETW needs to be able find WPT in non-default install locations


randomascii:
Some developers install WPT on drives other than C:. UIforETW either needs a way for the user to specify the location or it needs to automatically find the install location, perhaps by searching the PATH environment variable?

mwinterb:
HKLM\SOFTWARE\WOW6432Node\Microsoft\Microsoft SDKs\Windows\v8.1 and HKLM\SOFTWARE\WOW6432Node\Microsoft\Microsoft SDKs\Windows\v10.0 both have InstallationFolder values. They point to the directories that contain the Windows Performance Toolkit directory (including for one of my coworkers that encountered this issue).

However, there's another tool that may install WPT, right? That's not the SDK or the extracted MSI's? I would need to check if that tool does the same installation.

What is your preferred behavior if the registry locations don't exist? Fallback to the current "assume in Program Files"?

randomascii:
I'm not aware of a way to install WPT other than the SDK or the redist MSIs, although there could be such a thing. Falling back to the current behavior is probably fine if the registry keys don't exist or if xperf.exe can't be found.

I think there is also a problem on first-use of UIforETW where if it installs WPT and then tries to launch it then xperf.exe may not be found because it is added to the path too late. I have not seen this for a while. It resolves on relaunch of UIforETW.

randomascii:
Is this still an issue? Can I close this?

eugeneia:
FWIW I still had this issue even when setting the registry key and ended up using an NTFS junction to get it to work.

rickbrew:
> I'm not aware of a way to install WPT other than the SDK or the redist MSIs

You can also install it from the Microsoft Store, but I'm not sure if it's been maintained: https://www.microsoft.com/en-us/p/windows-performance-analyzer/9n0w1b2bxgnz?activetab=pivot:overviewtab

Detecting the app can be accomplished with something like [`GetPackagesByPackageFamily`](https://docs.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getpackagesbypackagefamily), and launching the app can be done with [`IApplicationActivationManager::ActivateApplication`](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-iapplicationactivationmanager-activateapplication). Powershlle's `Get-AppxPackage` is also useful for stumbling around and exploring this area.

randomascii:
[...]

PR #119, opened 2018-05-18
Added support for WPT installation in atypical locations.

mwinterb:
Addresses #118.

### DIFF
--- a/UIforETW/Utility.cpp
+++ b/UIforETW/Utility.cpp
@@ -298,6 +298,47 @@ std::wstring ConvertToCRLF(const std::wstring& input)
 	return result;
 }
 
+std::wstring ReadRegistryString(HKEY root, const std::wstring& subkey, const std::wstring& valueName, bool force32Bit)
+{
+	std::wstring value;
+	const DWORD flags = RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ | RRF_ZEROONFAILURE;
+
+	REGSAM openOptions = KEY_QUERY_VALUE;
+	if (force32Bit)
+	{
+		openOptions |= KEY_WOW64_32KEY;
+	}
+
+	HKEY key;
+	if (::RegOpenKeyExW(root, subkey.c_str(), 0, openOptions, &key) != ERROR_SUCCESS)
+	{
+		return value;
+	}
+
+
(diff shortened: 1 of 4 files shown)


UIforETW/Utility.cpp
```diff
@@ -298,6 +298,47 @@ std::wstring ConvertToCRLF(const std::wstring& input)
...
+	}
+
+
+	DWORD bufSize = 50 * sizeof(wchar_t);
+	LSTATUS result = ERROR_MORE_DATA;
+	while (result == ERROR_MORE_DATA)
+	{
+		value.resize(bufSize / sizeof(wchar_t));
+		DWORD type;
+		result = ::RegGetValueW(key, nullptr, valueName.c_str(), flags, &type, const_cast<wchar_t*>(value.data()), &bufSize);
```
mwinterb: This introduces a new CoreGuidelines analysis warning about using `const_cast`. It's not needed in C++17 (or latest) mode, but I didn't feel comfortable changing the build requirements for the project. But since *eventually* the `const_cast` isn't needed, it felt okay to do.

randomascii:
Thanks! Changing the build requirements seems reasonable but is best done in a separate change.

UIforETW/Utility.cpp
```diff
@@ -298,6 +298,47 @@ std::wstring ConvertToCRLF(const std::wstring& input)
 	return result;
 }
 
+std::wstring ReadRegistryString(HKEY root, const std::wstring& subkey, const std::wstring& valueName, bool force32Bit)
+{
+	std::wstring value;
+	const DWORD flags = RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ | RRF_ZEROONFAILURE;
```
firedaemon-cto, line 304: RegGetValue doesn't like RRF_RT_REG_EXPAND_SZ w/o RRF_NOEXPAND. Since this is then of little value, just remove RRF_RT_REG_EXPAND_SZ?

trueqbit, line 304: RegGetValue doesn't like RRF_RT_REG_EXPAND_SZ w/o RRF_NO_EXPAND.
I guess RRF_NO_EXPAND is of little value, hence just remove RRF_RT_REG_EXPAND_SZ?

PR #122, opened 2018-08-27
Update ProductName in UIforETW resource data

Qartar:
This field is visible in the Details tab of the executable's file
properties in Windows File Explorer. Other fields are filled out so
having "TODO: <Product name>" felt very out of place.


randomascii:
Thanks for the change - I'm not sure how I missed that one.

You can also add yourself to the CONTRIBUTORS and AUTHORS files as part of this change. I don't think it's required, but feel free to.

Qartar:
Thanks, Bruce. Feels a bit pretentious to add myself for a one-liner like this. Feel free to merge as-is.

randomascii:
Thanks for contributing!

PR #123, opened 2018-08-31
Changed to be able to support linking with lld-link.

mwinterb:
lld-link doesn't support the /manifestdependency coming from a #pragma comment instruction.
"lld-link.exe : error : /manifestdependency: is not allowed in .drectve"
https://github.com/llvm-mirror/lld/blob/bf86a304d1fee6d332c479bdb98cec1e44001f62/COFF/Driver.cpp#L284-L328
See https://bugs.llvm.org/show_bug.cgi?id=38797

The SupportsJustMyCode, GenerateDebugInformation, and LinkTimeCodeGeneration customizations all seem like reasonable changes to the LLVM.Cpp.Common.targets file, but this seems like a lower barrier of entry.
(The GenerateDebugInformation change is because 'true' gets translated to FastLink in the Microsoft msbuild files.)
See also https://bugs.llvm.org/show_bug.cgi?id=38799

### DIFF
--- a/UIforETW/UIforETW.vcxproj
+++ b/UIforETW/UIforETW.vcxproj
@@ -122,6 +122,7 @@
       <UACExecutionLevel>RequireAdministrator</UACExecutionLevel>
       <AdditionalLibraryDirectories>..\lib</AdditionalLibraryDirectories>
       <AdditionalDependencies>ETWProviders.lib</AdditionalDependencies>
+      <AdditionalManifestDependencies>"type='win32' name='Microsoft.Windows.Common-Controls' version='6.0.0.0' processorArchitecture='x86' publicKeyToken='6595b64144ccf1df' language='*'"</AdditionalManifestDependencies>
     </Link>
     <Midl>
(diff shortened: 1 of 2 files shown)


mwinterb:
(I may have an ulterior motive with this to use your clout to suggest that manifestdependency coming from .drectve is rather important since that is how the project templates enable themes by default.)

Also, since treat warnings as errors is enabled, and clang is generating more warnings, this doesn't actually compile cleanly out of the gate, but it at least nominally runs once that setting is removed.

randomascii:
Is there a bug filed to request lld-link support for this? If not you should file one and link to it from the PR description, and link to the PR from the bug.

It looks like this adds some support for clang-cl as well as lld-link, correct?

mwinterb:
Both issues have been reported to llvm. And I've updated the commit message to refer to those links.

mwinterb:
Oh, I thought I had made this comment: the JustMyCode setting was to avoid a warning, the others were hard errors, but that's why I had unintentionally omitted the mention from the description.

PR #130, opened 2019-01-19
Persist width and height of UIForETW

chrdavis:
This change implements the requested feature to persist the width and height of the UIForETW dialog.  Setting is read during startup and applied to the dialog window upon initialization.  I added a min size which is the default size of the dialog resource.  I also added some arbitrary large max values for the width and height.

### DIFF
--- a/UIforETW/UIforETWDlg.h
+++ b/UIforETW/UIforETWDlg.h
@@ -211,6 +211,13 @@ private:
 	int initialHeight_ = 0;
 	int lastWidth_ = 0;
 	int lastHeight_ = 0;
+	const int minWidth_ = 870;
+	const int minHeight_ = 398;
+	const int maxWidth_ = 3000;
+	const int maxHeight_ = 3000;
+	// Width and height persisted to settings
+	int previousWidth_ = 0;
+	int previousHeight_ = 0;
 
 	void SetSymbolPath();
 	// Call this to retrieve a directory from an environment variable, or use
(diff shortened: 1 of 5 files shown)


UIforETW/UIforETWDlg.h
```diff
@@ -211,6 +211,13 @@ class CUIforETWDlg : public CDialog
 	int initialHeight_ = 0;
 	int lastWidth_ = 0;
 	int lastHeight_ = 0;
+	const int minWidth_ = 870;
```
randomascii: Can these be set to zero? Then the initialWidth_ checks would handle the minimum. Having hard-coded minimums in the source seems undesirable because the minimum depends on system settings and will change after future dialog edits.

UIforETW/UIforETWDlg.h
```diff
@@ -211,6 +211,13 @@ class CUIforETWDlg : public CDialog
 	int initialHeight_ = 0;
 	int lastWidth_ = 0;
 	int lastHeight_ = 0;
+	const int minWidth_ = 870;
+	const int minHeight_ = 398;
+	const int maxWidth_ = 3000;
```
randomascii: The same comments apply to maxWidth_ although setting them to large numbers makes this mostly academic, so I'm fine with leaving them as is.

randomascii:
I left some comments on UIforETWDlg.h

UIforETW/UIforETWDlg.h
chrdavis: Agreed.  Updated.

chrdavis:
I wasn't sure if the feature request also wanted to preserve the position as well as the width and height.  Would require added code to handle checking valid coordinates and potential scale changes.

randomascii:
Seems to work. Thanks for the contribution. Remembering position seems less important - it's often better to let Windows do its default behavior.

PR #138, opened 2019-11-29
Fix the locale dependent date parsing used for the output heap trace …

lukastvrdy:
…filename

date/T prints the date in locale-dependent format and thus causes issues when being parsed.
wmic command works independently of OS locale, language or the user's chosen date format.


lukastvrdy:
@googlebot I signed it!

randomascii:
Sorry for the delayed response - I went on vacation the day you uploaded this. I'm hitting some problems with the find command but I think that's just because I've got a non-standard find command in my path. I will look more closely later.

randomascii:
The find problem was unrelated. Merged

PR #148, opened 2021-02-08
Fix poor detection of a dupe instance

### LINKED ISSUE #147: UIforETW shortcut doesn't launch from within a folder named 'UI for ETW'
A `UIforETW.lnk` shortcut created from a `...\etwpackage\bin\UIforETW.exe` file doesn't lanch the UIforETW app if this shortcut is placed in a folder named `UI for ETW` — the shortcut fails right after asking for elevated permissions.
The same shortcut placed in a different folder (or if the same folder is renamed) works just fine

Windows 10.0.19042.0
etwpackage1.54.zip

eugenesvk:
Old FindWindow() function could mistake an Explorer folder 'UI for ETW'
for another instance of 'UI for ETW' (since window title is identical)
New custom function registers a unique window message that allows to
uniquely identify another instance of 'UI for ETW'
Bonus: restore the minimized window of 'UI for ETW'

Resolves: #147

### DIFF
--- a/UIforETW/UIforETW.cpp
+++ b/UIforETW/UIforETW.cpp
@@ -170,6 +170,25 @@ CUIforETWApp::CUIforETWApp() noexcept
 
 CUIforETWApp theApp;
 
+BOOL CALLBACK CUIforETWApp::findDupeWindow(HWND hWnd, LPARAM lParam)
+{
+	DWORD_PTR result;
+	LRESULT msgRet = SendMessageTimeout(hWnd, uwmAreYouMe, 0, 0,
+		SMTO_BLOCK |      // Don't process any other requests until the function returns
+		SMTO_ABORTIFHUNG, // No waiting if the thread appears to not respond
+		200, &result);    // Timeout in ms
+		if(msgRet == 0)           // failed to send message to this window, ignore it
+		{
+			return TRUE;
+		}
+		if(result == uwmAreYouMe) // found a dupe window
+		{
+			HWND * target = (HWND *)lParam;
+			*target = hWnd;
+			return FALSE;           // stop window enumeration
+		}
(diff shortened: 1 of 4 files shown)


randomascii:
It looks basically good but I'll need the cla/google signed before I can accept the patch.

eugenesvk:
@googlebot I signed it!

randomascii:
I get four instances of this error when compiling:

1>C:\src\github\UIforETW\UIforETW\Settings.cpp(17,10): fatal error C1853: 'x64\Release\UIforETW_devrel.pch' precompiled header file is from a different version of the compiler, or the precompiled header is C++ and you are using it from C (or vice versa)

I don't mind switching to C++17 but I don't think that global constructors are a great idea. How about:

int GetAreYouMeId() {
  static const UINT uwmAreYouMe = RegisterWindowMessage(L"678fd291e05741edb987c0a4f5650866");
  return uwmAreYouMe;
}

Then it gets constructed on first use instead of before WinMain

randomascii:
The return type in my sample function should be UINT.

randomascii:
I noticed this was still open and decided to see why. Part of the reason is that I pasted the wrong error message into my last comment. The actual error message is:

1>C:\src\github\UIforETW\UIforETW\UIforETW.h(46,19): error C7525: inline variables require at least '/std:c++17' (compiling source file Settings.cpp)

That message happens on this line:

inline const UINT uwmAreYouMe = RegisterWindowMessage(L"678fd291e05741edb987c0a4f5650866");

I continue to think that that isn't a great idea. Inline data is great. Global constructors that run before main are not.

dneto0:
Sorry, this repo is now archived.

Further development has moved to https://github.com/randomascii/UIforETW

PR #149, opened 2021-03-01
Load registry value string only of type REG_SZ

trueqbit:
Amends #118


trueqbit:
@googlebot I signed it!

randomascii:
Thanks!

PR #167, opened 2024-02-01
Update the Chrome event provider name in the WPA profiles

sebmarchand:
The Chrome event provider name changed from "Chrome" to "Google.Chrome" in https://chromium-review.googlesource.com/c/chromium/src/+/2171974 .

Using the new name ensure that the preset WPA filters that come with UIForETW keeps working.


dneto0:
Sorry, this repo is now archived.

Further development has moved to https://github.com/randomascii/UIforETW

